PR #1, opened 2018-02-16
Fixed host and pathname handling

delvedor:
As titled.
Both fixes are tested in `test/host-header.js`.


delvedor:
Hold on, there is still a little bug.

delvedor:
This test is failing, any idea why?
https://github.com/fastify/fastify-reply-from/blob/661a7835f35143ff3204f21e2d1092c31827e6d6/test/host-header.js#L24-L29

PR #2, opened 2018-02-21
Add http2 support

mcollina:


mcollina:
no it doesn't. It's just http2 -> http.

PR #6, opened 2018-04-21
Added support for http2 target

### LINKED ISSUE #5: http2 support for target server
I would like to use this as proxy for http2 target fastify server. But right now it seems that it does not support target as http2.
If you could provide guidance I will provide pull request for http2 support and request id tracking.
Please let me know.

nileshmali:
Closes #5

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # fastify-reply-from
 
+[![Greenkeeper badge](https://badges.greenkeeper.io/fastify/fastify-reply-from.svg)](https://greenkeeper.io/)
+
 [![Build Status](https://travis-ci.org/fastify/fastify-reply-from.svg?branch=master)](https://travis-ci.org/fastify/fastify-reply-from)
 [![NPM version](https://img.shields.io/npm/v/fastify-reply-from.svg?style=flat)](https://www.npmjs.com/package/fastify-reply-fromm)
 
--- a/README.md
+++ b/README.md
@@ -59,8 +59,11 @@ target.listen(3001, (err) => {
 
 #### `base`
 
-Set the base URL for all the forwarded requests.
+Set the base URL for all the forwarded requests. Will be required if `http2` is set to `true`
 Note that _every path will be discarded_.
+
+#### http2
+Set to `true` if target server is `http2` enabled.
 
 #### cacheURLs
 
@@ -96,7 +99,7 @@ Called when an http response is received from the source.
 The default behavior is `reply.send(res)`, which will be disabled if the
 option is specified.
 
-#### rewriteHeaders(res)
+#### rewriteHeaders(headers)
 
 Called to rewrite the headers of the response, before them being copied
 over to the outer response.
@@ -126,7 +129,6 @@ already overriding the [`body`][body].
 * [ ] forward the request id to the other peer might require some
       refacotring because we have to make the `req.id` unique
       (see [hyperid](http://npm.im/hyperid)).
-* [ ] support http2
 * [ ] benchmarks
 
 ## License
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ already overriding the [`body`][body].
 * [ ] forward the request id to the other peer might require some
       refacotring because we have to make the `req.id` unique
       (see [hyperid](http://npm.im/hyperid)).
+* [ ] Support origin HTTP2 push
 * [ ] benchmarks
 
 ## License
--- a/index.js
+++ b/index.js
@@ -1,28 +1,24 @@
 'use strict'
 
 const fp = require('fastify-plugin')
-const http = require('http')
-const https = require('https')
 const URL = require('url').URL
 const lru = require('tiny-lru')
 const querystring = require('querystring')
 const Stream = require('stream')
-const pump = require('pump')
-const requests = {
-  'http:': http,
-  'https:': https
-}
+const RequestAgent = require('./lib/request')
+const { filterPseudoHeaders, copyHeaders, stripHttp1ConnectionHeaders } = require('./lib/utils')
 
 module.exports = fp(function from (fastify, opts, next) {
-  const agents = {
-    // with a colon, so that it matches url.protocol
-    // and we can avoid string manipulation at runtime
-    'http:': new http.Agent(agentOption(opts)),
-    'https:': new https.Agent(agentOption(opts))
-  }
   const cache = lru(opts.cacheURLs || 100)
   const base = opts.base
-
+  const requestClient = new RequestAgent({
+    http2: !!opts.http2,
+    base,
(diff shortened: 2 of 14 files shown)


nileshmali:
How do I prevent http2 tests running for Node 8 on travis?
Any hints?

mcollina:
Wih something like https://github.com/fastify/fastify/blob/master/test/http2/http2.test.js

lib/request.js
```diff
@@ -0,0 +1,102 @@
+'use strict'
+const http = require('http')
+const https = require('https')
+const http2 = require('http2')
+const eos = require('end-of-stream')
+const pump = require('pump')
+const { stripHttp1ConnectionHeaders } = require('./utils')
+
+class RequestAgent {
```
mcollina: I'm very -1 on having this as a class. Can you restore the approach with just functions?

Something like:


```js
function buildRequest (opts) {
  return { request, close }

  function request () {}
  function close () {}
}
```

lib/request.js
```diff
@@ -0,0 +1,102 @@
...
+  constructor (opts) {
+    this.http2 = !!opts.http2
+    this.requests = {
+      'http:': http,
+      'https:': https
+    }
+    if (this.http2) {
+      if (!opts.base) throw new Error('Option base is required when http2 is true')
```
mcollina: Nice one!

lib/request.js
```diff
@@ -0,0 +1,102 @@
...
+    })
+    req.on('error', done)
+    req.on('response', (res) => {
+      done(null, { statusCode: res.statusCode, headers: res.headers, stream: res })
+    })
+    end(req, body, done)
+  }
+
+  handleHttp2Req ({ method, url, qs, headers, body }, done) {
+    this.http2Client.ref()
```
mcollina: why is this `ref()` needed?

README.md
```diff
@@ -126,7 +129,6 @@ already overriding the [`body`][body].
 * [ ] forward the request id to the other peer might require some
       refacotring because we have to make the `req.id` unique
       (see [hyperid](http://npm.im/hyperid)).
-* [ ] support http2
 * [ ] benchmarks
```
mcollina: Can you add another todo? Support origin HTTP2 push.

test/http2-http2.js
```diff
@@ -0,0 +1,61 @@
...
+const h2url = require('h2url')
+const t = require('tap')
+const Fastify = require('fastify')
+const From = require('..')
+const fs = require('fs')
+const path = require('path')
+const certs = {
+  key: fs.readFileSync(path.join(__dirname, 'fixtures', 'fastify.key')),
+  cert: fs.readFileSync(path.join(__dirname, 'fixtures', 'fastify.cert'))
+}
```
mcollina: Can you add a test from http (non-secure) to http2 (non-secure)?

lib/request.js
```diff
@@ -0,0 +1,102 @@
...
+    const req = this.http2Client.request({
+      ':method': method,
+      ':path': url.pathname + qs,
+      ...stripHttp1ConnectionHeaders(headers)
+    })
+    if (!isGet) {
+      end(req, body, done)
+    }
+    req.on('error', done)
+    eos(req, () => { this.http2Client.unref() })
```
mcollina: Why is `unref()` needed?

lib/utils.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+  return dest
+}
+
+function copyHeaders (headers, reply) {
+  const headersKeys = Object.keys(headers)
+
+  let header
+
+  for (let i = 0; i < headersKeys.length; i++) {
```
mcollina: this use of let is __very__ slow on Node 8. Can you please restore `var` everywhere?

index.js
```diff
@@ -55,10 +51,11 @@ module.exports = fp(function from (fastify, opts, next) {
         opts.contentType = 'application/json'
       }
 
-      headers = Object.assign(headers, {
+      headers = {
+        ...headers,
```
delvedor: As far as I know use a plain old `for` loop is way faster than the spread operator to copy object keys and values.
Since this is a hot path you should probably use the `for` loop.
cc @mcollina

lib/request.js
```diff
@@ -0,0 +1,102 @@
...
+      this.http2Client = http2.connect(opts.base, { rejectUnauthorized: opts.rejectUnauthorized })
+    } else {
+      this.agents = {
+        'http:': new http.Agent(agentOption(opts)),
+        'https:': new https.Agent(agentOption(opts))
+      }
+    }
+  }
+
+  handleHttp1Req ({ method, url, qs, headers, body }, done) {
```
delvedor: The deconstructing assignment is slower than the classic `obj.key` assignment. Can you use it?

lib/request.js
nileshmali: https://github.com/mcollina/h2url/blob/c23d1e920308f68ef2ba5d49720adc54326dd1b5/index.js#L28
I went by the comment on above. If not needed I will remove it. May be I mis-interpreted.

test/http2-http2.js
nileshmali: Already there in [http-http2.js](https://github.com/fastify/fastify-reply-from/pull/6/files#diff-744c3c82c325d56497f72b3780a5ef4b) file

lib/request.js
```diff
@@ -0,0 +1,104 @@
...
+    })
+    req.on('error', done)
+    req.on('response', (res) => {
+      done(null, { statusCode: res.statusCode, headers: res.headers, stream: res })
+    })
+    end(req, opts.body, done)
+  }
+
+  function handleHttp2Req (opts, done) {
+    http2Client.ref()
```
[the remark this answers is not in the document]
nileshmali: I will remove call to .ref() and .unref() as we are explicitly destroying client in .onClose()

index.js
nileshmali: I will change it to:
```javascript
headers['content-length'] = Buffer.byteLength(body)
headers['content-type'] = opts.contentType
```

delvedor: Why not just copy the actual headers?

nileshmali: Actual headers are already copied. So append is enough.

nileshmali:
@mcollina and @delvedor Thanks for the review. I will submit updated code by tomorrow.

lib/request.js
```diff
@@ -0,0 +1,104 @@
...
+function buildRequest (opts) {
+  const isHttp2 = !!opts.http2
+  const requests = {
+    'http:': http,
+    'https:': https
+  }
+  var http2Client
+  var agents
+
+  if (isHttp2) {
```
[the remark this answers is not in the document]
nileshmali: Done

mcollina:
Done!

PR #9, opened 2018-05-02
Fix crash when target server is unavailable

### LINKED ISSUE #8: Crash: Source crashes when http2 target goes down or becomes unavailable
**Scenario 1:** 
Start source server without starting client. Source process crashes immediatly as it tries to create client connection eagerly.

**Scenario 2:**
1. Start target and then source 
2. Shutdown target service
3. Hit proxied request
Source server crashes as http2 client session becomes invalid after target shutdown

**Proposed Solution:**
1. Lazy initialization of `http2Client` on first request.
2. Setup `error` event listener on `http2Client`.
Planning to return status `500` with `Internal Server Error` as message. **_Any suggestions?_**
3. Check `http2Client.destroyed` flag to determine whether session is valid or not and attempt to reconnect.

Pull request will follow shortly.

nileshmali:
Fixes #8

### DIFF
--- a/index.js
+++ b/index.js
@@ -66,7 +66,13 @@ module.exports = fp(function from (fastify, opts, next) {
     request({ method: req.method, url, qs, headers, body }, (err, res) => {
       if (err) {
         this.request.log.warn(err, 'response errored')
-        this.send(err)
+        if (!this.sent) {
+          if (err.code === 'ERR_HTTP2_STREAM_CANCEL') {
+            this.code(503).send(new Error('Service Unavailable'))
+          } else {
+            this.send(err)
+          }
+        }
         return
       }
       this.request.log.info('response received')
--- a/lib/request.js
+++ b/lib/request.js
@@ -13,6 +13,8 @@ function buildRequest (opts) {
     'http:': http,
     'https:': https
   }
+  const baseUrl = opts.base
+  const rejectUnauthorized = opts.rejectUnauthorized
   var http2Client
   var agents
 
@@ -21,7 +23,6 @@ function buildRequest (opts) {
       throw new Error('Http2 support requires Node version >= 9.0.0')
     }
     if (!opts.base) throw new Error('Option base is required when http2 is true')
-    http2Client = http2.connect(opts.base, { rejectUnauthorized: opts.rejectUnauthorized })
   } else {
     agents = {
       'http:': new http.Agent(agentOption(opts)),
@@ -34,7 +35,7 @@ function buildRequest (opts) {
     if (!isHttp2) {
       agents['http:'].destroy()
       agents['https:'].destroy()
-    } else {
+    } else if (http2Client) {
       http2Client.destroy()
     }
   }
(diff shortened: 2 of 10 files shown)


index.js
```diff
@@ -66,7 +66,13 @@ module.exports = fp(function from (fastify, opts, next) {
     request({ method: req.method, url, qs, headers, body }, (err, res) => {
       if (err) {
         this.request.log.warn(err, 'response errored')
-        this.send(err)
+        if (!this.sent) {
+          if (err.code === 'ERR_HTTP2_STREAM_CANCEL') {
```
nileshmali: Not sure whether this is correct way or not. I am trying to avoid exposing internal error message which contains host address information of target.
Please suggest.

nileshmali: Error is already logged on line 68. Which should have all information.

lib/request.js
```diff
@@ -56,12 +57,16 @@ function buildRequest (opts) {
   }
 
   function handleHttp2Req (opts, done) {
-    const isGet = opts.method === 'GET' || opts.method === 'get'
+    if (!http2Client || http2Client.destroyed) {
+      http2Client = http2.connect(baseUrl, { rejectUnauthorized })
+      http2Client.once('error', done)
```
mcollina: this `done` call should be removed once the new client is connected.

mcollina:
Can you please check if you have the latest version of avvio installed? We had a bug on a similar topic.

nileshmali:
avvio version is `5.4.3` which is latest. Still have same result.

mcollina:
You have to `await server.close()` in both occasions.

nileshmali:
Seems like `server.close` does not return promise, which causes timeout error.

mcollina:
I'm pretty certain that it does.

nileshmali:
Yeah! You were right. Issue is with `http2` flag on target.
If you remove `got` calls `server.close()` works without any issue. But if you fire any request then `server.close()` does not work and times out. 
I suspect that, as Http2Server sessions are persistent they are still active after `got` request and hence preventing graceful shutdown of fastify server.

Is this related to below note from NodeJS documentation:
> Note that this is not analogous to restricting new requests since HTTP/2 connections are persistent. To achieve a similar graceful shutdown behavior, consider also using http2session.close() on active sessions.

mcollina:
Part of the problem is that in https://github.com/nileshmali/fastify-reply-from/blob/21ec5dbb7130204d6c9065c54f32b19010138254/index.js#L94-L97 we don not wait for things to be shut down properly.

nileshmali:
But I am not trying to close instance which has plugin registered. I am closing target instance which does not have any of the plugin installed.

mcollina:
Can you please add the failing test to this PR? It'd be easier for me to help.

nileshmali:
@mcollina Pushed failing test file, please take a look.

mcollina:
It ended up being a node core bug on Http2: https://github.com/nodejs/node/issues/20630. The callback we pass to `close(cb)` is never called.

mcollina:
Can you please remove the test?

nileshmali:
Done

mcollina:
Thanks @nileshmali! Would you like to help maintaining this?

nileshmali:
Yes. I would be glad to do that. But have very less time during weekdays. 
My responses might be delayed.

PR #12, opened 2018-05-27
reset request body when proxy get request from fastify

vincent178:
according to https://tools.ietf.org/html/rfc2616#section-4.3. fastify ignore message body when it's a GET request. when proxy this request, we should reset the content-length to make it a valid http request discussion: https://github.com/fastify/fastify/issues/953

### DIFF
--- a/index.js
+++ b/index.js
@@ -66,7 +66,7 @@ module.exports = fp(function from (fastify, opts, next) {
     // when proxy this request, we should reset the content-length to make it a valid http request
     // discussion: https://github.com/fastify/fastify/issues/953
     if (req.method === 'GET') {
-      headers['content-length'] = Buffer.byteLength(body)
+      headers['content-length'] = 0
     }
 
     req.log.info({ source }, 'fetching from remote server')
--- a/index.js
+++ b/index.js
@@ -61,6 +61,14 @@ module.exports = fp(function from (fastify, opts, next) {
       }
     }
 
+    // according to https://tools.ietf.org/html/rfc2616#section-4.3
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -61,6 +61,14 @@ module.exports = fp(function from (fastify, opts, next) {
       }
     }
 
+    // according to https://tools.ietf.org/html/rfc2616#section-4.3
+    // fastify ignore message body when it's a GET request
+    // when proxy this request, we should reset the content-length to make it a valid http request
+    // discussion: https://github.com/fastify/fastify/issues/953
+    if (req.method === 'GET') {
+      headers['content-length'] = Buffer.byteLength(body)
```
mcollina: I would reset it to zero. GETs do not have bodies, and we should not forward them.

vincent178:
Sure, my original thought is future compatibility if we want to use the body in some way. But I agree with you, we don't need think that way. I will change it.

vincent178:
updated @mcollina

index.js
```diff
@@ -61,6 +61,14 @@ module.exports = fp(function from (fastify, opts, next) {
       }
     }
 
+    // according to https://tools.ietf.org/html/rfc2616#section-4.3
+    // fastify ignore message body when it's a GET request
+    // when proxy this request, we should reset the content-length to make it a valid http request
+    // discussion: https://github.com/fastify/fastify/issues/953
+    if (req.method === 'GET') {
```
delvedor: We should do the same for `HEAD` requests.

PR #13, opened 2018-05-28
Be more strict when rewriting a body

mcollina:


index.js
```diff
@@ -65,7 +65,10 @@ module.exports = fp(function from (fastify, opts, next) {
     // fastify ignore message body when it's a GET request
     // when proxy this request, we should reset the content-length to make it a valid http request
     // discussion: https://github.com/fastify/fastify/issues/953
-    if (req.method === 'GET') {
+    if (req.method === 'GET' || req.method === 'HEAD') {
+      if (body) {
+        throw new Error('Rewriting the body when doing a GET is not allowed')
```
delvedor: With this we will throw at runtime.

mcollina: This is expected. It will happen only if `opts.body`  is specified (`this.request.body`  will be null if not). So it's safe and will prevent an error.

Do you want me to add a comment?

delvedor: > Do you want me to add a comment?

delvedor: Probably you can even do a more strict check, with `body == null`

mcollina: done

index.js
```diff
@@ -65,7 +65,13 @@ module.exports = fp(function from (fastify, opts, next) {
     // fastify ignore message body when it's a GET request
```
vincent178: it would be better to update the comment as well

vincent178:
nice

PR #40, opened 2019-01-07
Fix incorrect content-length and content-type for padded bodies

wei-liddat:
If an incoming request has a JSON-stringified body with non-zero spacing, the content length will be incorrect after parsing/stringifying it.

Since the headers are just forwarded, the content-length header will have an incorrect number of bytes, and will result in the target server either hanging when waiting for more bytes, or truncating the body if there are less bytes than specified.

This recalculates the length of the body, and sets the type as well (this should be json since it is stringifed by JSON, right?)

### DIFF
--- a/index.js
+++ b/index.js
@@ -59,6 +59,8 @@ module.exports = fp(function from (fastify, opts, next) {
         body = this.request.body
       } else {
         body = JSON.stringify(this.request.body)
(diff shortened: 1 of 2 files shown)


cemremengu:
@wei-liddat landed in ^0.5.4

PR #52, opened 2019-04-27
Allow overwrite request client

jkyberneees:
Optionally allow the `request http client` overwrite in `from` call.
This extension will allow to proxy requests to different target APIs using the same plugin instance. 

Example use case: https://www.npmjs.com/package/k-fastify-gateway

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,7 +15,7 @@ const {
 module.exports = fp(function from (fastify, opts, next) {
   const cache = lru(opts.cacheURLs || 100)
   const base = opts.base
-  const { request, close } = buildRequest({
+  const { request: defaultRequest, close } = buildRequest({
     http2: !!opts.http2,
     base,
     keepAliveMsecs: opts.keepAliveMsecs,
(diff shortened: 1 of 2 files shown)


mcollina:
I’m somewhat hesitant to have the _internal_ request object exposed. I’ll rather prefer we introduce a new kind of API to make the creation of the request/connection pool dynamic or swappable at runtime.

What’s exactly your use case, how do you plan to change/use this at runtime?

jkyberneees:
My exact use case would be to support services with different settings behind the gateway, for example:
```js
const fastify = require('fastify')({})
fastify.register(require('fastify-reply-from'))
fastify.register(require('k-fastify-gateway'), {

  routes: [{
    prefix: '/public',
    target: 'http://lb-public-service:3000'
  }, {
    prefix: '/admin',
    target: {
      base: 'https://lb-admin-service:3000',
      http2: true
    }
  }]
})

fastify.listen(8080).then((address) => {
  console.log(`API Gateway listening on ${address}`)
})
```
Here I plan to maintain and inject different `request` instances for each target service. 

However, if this features are built-in inside `fastify-reply-from` -> `from` method, it would be also great.

mcollina:
You can already do that, just register fastify-reply-from under two different prefixes.

jkyberneees:
You are right. Will use that approach instead.
I let to your consideration the closing of this PR. 
Thanks.

PR #53, opened 2019-04-29
add options to specify certificates in plugin registration or into the decorator 'from'

matth-c3:
Hello @mcollina ,

added the options `cert` and `key` in the `opts` of the plugin, and in the `opts` of the decorator `from`. 
To be more specific you can declare the certificate during the fastify.register():
```
fastify.register(require('npm-custom/fastify-reply-from'), {
   certificate: {
      cert: params.cert,
      key: params.key
   }
})
```
or directly in the from declaration as an option. It will create a "temp" `agent` with the certificate define in `opts`.
```
reply.from(`https:/${request.req.url}`, {
   certificate: {
      cert: params.cert,
      key: params.key
   }
})
```
In my case I use a different certificates depends on the `base` / `endpoint` so I declare them in the `from` directly.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -31,6 +31,12 @@ function buildRequest (opts) {
       'http:': new http.Agent(agentOption(opts)),
       'https:': new https.Agent(agentOption(opts))
     }
+    // if (opts.certificate) {
+    //   if (Buffer.isBuffer(opts.certificate.key) && Buffer.isBuffer(opts.certificate.cert)) {
+    //     agents['https'].cert = opts.certificate.cert
+    //     agents['https'].key = opts.certificate.key
+    //   }
+    // }
   }
 
   if (isHttp2) {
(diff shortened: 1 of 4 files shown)


lib/request.js
```diff
@@ -31,6 +31,12 @@ function buildRequest (opts) {
       'http:': new http.Agent(agentOption(opts)),
       'https:': new https.Agent(agentOption(opts))
     }
+    // if (opts.certificate) {
```
jsumners: Why is this commented out?

lib/request.js
```diff
@@ -58,13 +64,21 @@ function buildRequest (opts) {
   }
 
   function handleHttp1Req (opts, done) {
+    if (opts.certificate) {
+      agents['clone'] = new https.Agent(agentOption({
+        ...opts, ...opts.certificate
+      }))
+      // if (Buffer.isBuffer(opts.certificate.key) && Buffer.isBuffer(opts.certificate.cert)) {
```
jsumners: Why is this commented out?

mcollina:
Would you mind removing `package-lock.json` and adding a unit test? This needs to work on http1, http2, and http1 with undici.

lib/request.js
matth-c3: Hello @jsumners , 

The check is done directly in the function `agentOption` directly.

lib/request.js
matth-c3: Hello @jsumners , 

I was not sure if you want to check if `cert` and `key` are ` Buffer` or not, let me know if you want me to uncomment it.

lib/request.js
jsumners: If the code path is not useful, remove it.

matth-c3: if you mean `opts.certificate....` I use it in the `handleHttp1Req` function. I prefer to check if `opt.certificate` exist rather than both keys `opts.cert` and `opts.key`.

lib/request.js
matth-c3: Hello @jsumners , let me know if you want me to leave the `buffer` test and if you want me to throw an exception.

matth-c3:
Hello @mcollina ,

Do you mean that I should add an exception when it's `opts.certificate` is given when `undici` is use ?

mcollina:
I would like to see tests added that verifies that this is working with http2 and undici.

matth-c3:
Sorry I don't use those also in undici readme it's not mention that you can specify a certificat in the request.

Issue #59, opened 2019-05-31
Always sending requests `application/json` if not a stream


jsumners:
https://github.com/fastify/fastify-reply-from/blob/97508b89512d07b751bcef3d90ab2441343b1465/index.js#L65-L69

If a `POST` request is sent with a content type of `text/plain` and a body of `this is plain text` then the above branch will be hit and the outgoing request will have the content type header set to `application/json` and the request body will be JSON stringified.

1. The body should not be JSON stringified unless the received payload is designated `application/json`
2. It should pass along the received content type header as-is instead of rewriting it to `application/json`

mcollina:
LOL, this is nice. Would you like to send a PR?

jsumners:
I don't have permission to push my branch up to this repo.

mcollina:
now you do

Il giorno sab 1 giu 2019 alle ore 14:24 James Sumners <
notifications@github.com> ha scritto:

mcollina:
Closed in #60.

PR #60, opened 2019-06-03
Resolves issue #59

jsumners:
As titled. This PR resolved issue #59.

I am not clear why the body needs to be stringified if it is `application/json`. Shouldn't it just be passed along to the remote host?


jkyberneees:
JFYI: https://github.com/fastify/fastify-reply-from/issues/61

Issue #61, opened 2019-06-03
Invalid content-type check


jkyberneees:
Hi team, I wanted to report that the following statement:
https://github.com/fastify/fastify-reply-from/blob/4e8718d86443ddb2a0c53c0c629bc8eb6460a315/index.js#L68
Will fail in case of extended content-type formats like: `application/json;charset=utf-8`

Check should be:
```js
body = contentType.toLowerCase().indexOf('application/json') === 0 ? JSON.stringify(this.request.body) : this.request.body
```

Can we fix this one? 
Regards

jkyberneees:
ping @jsumners

jsumners:
@jkyberneees there's no way I'd be able to get to it before Saturday. A PR for `.startsWith()` would be welcome.

PR #62, opened 2019-06-04
considering extended content-type header format

jkyberneees:
Fixes: https://github.com/fastify/fastify-reply-from/issues/61


mcollina:
Can you please add a unit test?

jkyberneees:
Sure, ASAP

PR #66, opened 2019-12-18
Add support for unix socket destination server

coreyfarrell:
This is not ready for merge as tests / documentation are needed, at this time I'm really just looking for feedback on the proposed API.  I can confirm `unix+http` works, I've used this branch to replace the `fastify-reply-from` used by `fastify-http-proxy`, the following is an example initialization:
```js
fastify.register(fastifyHttpProxy, {
  upstream: `unix+http://${querystring.escape(runDir('http.sock'))}/`,
  prefix: this.prefix
});
```

I got the idea of encoding the path as hostname from a [CURL thread](https://curl.haxx.se/mail/lib-2008-04/0280.html).  The separate `unix+http` protocol is easier to detect and should not impose a performance cost on TCP hostname based requests.  Also `unix+http` avoids potential ambiguity if someone uses a socket file in `cwd` such as `unix+http://file.sock/resource`, this will not attempt to load from the `file.sock` hostname.

Worth noting I have not tested the `unix+https` functionality.  I'm not sure this is even possible and if it is I'm not convinced it's useful.  I'd be open to dropping `unix+https` from the initial patch, let someone else add that if they find a use for it.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,17 @@ target.listen(3001, (err) => {
 Set the base URL for all the forwarded requests. Will be required if `http2` is set to `true`
 Note that _every path will be discarded_.
 
+Custom URL protocols `unix+http:` and `unix+https:` can be used to forward requests to a unix
+socket server by using `querystring.escape(socketPath)` as the hostname.  This is not supported
+for http2 nor unidici.  To illustrate:
+
+```js
+const socketPath = require('querystring').escape('/run/http-daemon.socket')
+proxxy.register(require('fastify-reply-from'), {
+  base: 'unix+http://${socketPath}/'
+});
+```
+
 #### `http`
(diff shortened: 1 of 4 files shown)


mcollina:
I'm good with the approach, however this will require unit tests and docs.

coreyfarrell:
@mcollina sorry this got delayed so long but I think it's ready for review.  I've added tests and done my best to document (documentation is really not one of my strengths).

The plugin will throw an error for http2 and undici if the base URL starts with `unix+` (I'm not sure if/how those can be supported).  I did not add tests for these new error conditions, in my defense no tests exist for existing error (EX lack of `opts.base` for http2).  Maybe we should add a `test/errors.js` that would verify proper throws for invalid options?

PR #67, opened 2020-01-19
Updated dependencies, dropped old nodes and added sessiontTimeout option

mcollina:
Updated. I also added a new option, `sessionTimeout` because otherwise a test now would take 2 minutes.

#### Checklist


mcollina:
I would just bump the major here. There are quite a few changes that are semver-major anyway in the way it handles http2.

PR #70, opened 2020-03-28
Fix info of undefined (req.log.info) while modifyCoreObjects is set t…

g45t345rt:
…o false

#### Checklist

- [ ] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)


g45t345rt:
Good idea! 
As I suspected it does indeed fail when I instantiate a Fastify server with { modifyCoreObjects: false }

PR #74, opened 2020-04-23
feat: added onError

TriPSs:
This add's `onError`, behavior is the same as `onResponse` only this will be triggered when an error occurs.  

#### Checklist


index.js
```diff
@@ -103,7 +104,9 @@ module.exports = fp(function from (fastify, opts, next) {
       if (err) {
         this.request.log.warn(err, 'response errored')
         if (!this.sent) {
-          if (err.code === 'ERR_HTTP2_STREAM_CANCEL') {
+          if (onError) {
+            onError(this.request.req, this, err)
```
jsumners, line 108: I believe `this` will be an instance of Fastify's reply object. Thus, `this.request` should be passed instead of `this.request.req` so that a Fastify request object is passed.

TriPSs: `this.request.req` is Fastify's reply object, i also wanted to keep it aligned with `onResponse`.

jsumners: Oh. These are from the `request` library.

mcollina:
Implemented in #96.

PR #76, opened 2020-04-24
Support modifying of logging

markrzen:
I am unable to get the tests to run positively before I made this commit, but happy to work through it if the PR is generally well accepted?

```
 FAIL  test/http2-invalid-target.js
 ✖ Protocol error

  test: http -> http2
  at:
    line: 764
    column: 26
    file: internal/http2/core.js
    function: Http2Session.onSessionInternalError
  code: ERR_HTTP2_ERROR
  errno: -505
  tapCaught: uncaughtException
```

#### Checklist

- [ ] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,18 @@ Setting this option will not verify if the http method allows for a body.
 Override the `'Content-Type'` header of the forwarded request, if we are
 already overriding the [`body`][body].
 
+#### `logProxyError(err)`
+
+Override the default logging of proxy errors, allowing the changing of format or addition of information. You may also disable by setting to `false`.
+
+#### `logProxyRequest(source, requestConfig)`
+
+Override the default logging of proxy requests, allowing the changing of format or addition of information. You may also disable by setting to `false`.
+
+#### `logProxyResponse(res)`
+
+Override the default logging of proxy responses, allowing the changing of format or addition of information. You may also disable by setting to `false`.
+
 ### HTTP & HTTP2 timeouts
 
 This library has:
--- a/README.md
+++ b/README.md
@@ -199,15 +199,15 @@ Setting this option will not verify if the http method allows for a body.
 Override the `'Content-Type'` header of the forwarded request, if we are
 already overriding the [`body`][body].
 
-#### `logProxyError(err)`
+#### `logProxyError(log, err)`
 
 Override the default logging of proxy errors, allowing the changing of format or addition of information. You may also disable by setting to `false`.
 
-#### `logProxyRequest(source, requestConfig)`
+#### `logProxyRequest(log, source, requestConfig)`
 
 Override the default logging of proxy requests, allowing the changing of format or addition of information. You may also disable by setting to `false`.
 
-#### `logProxyResponse(res)`
+#### `logProxyResponse(log, res)`
 
 Override the default logging of proxy responses, allowing the changing of format or addition of information. You may also disable by setting to `false`.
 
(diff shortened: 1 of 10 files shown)


index.js
```diff
@@ -95,13 +99,19 @@ module.exports = fp(function from (fastify, opts, next) {
...
 
-    this.request.log.info({ source }, 'fetching from remote server')
-
     const requestHeaders = rewriteRequestHeaders(req, headers)
 
-    request({ method: req.method, url, qs, headers: requestHeaders, body }, (err, res) => {
+    const requestConfig = { method: req.method, url, qs, headers: requestHeaders, body }
+
+    if (logProxyRequest) {
+      logProxyRequest.apply(this, [source, requestConfig])
```
jsumners, line 107: The docs should indicate that the function is invoked with `this` bound to the Fastify reply object. And that `this.log` should be used to actually log things within the function.

markrzen:
@mcollina Happy to add some unit tests. Did you see my note above about the tests not running currently? Is this a common problem to the project, or something new?

index.js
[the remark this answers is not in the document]
markrzen: @mcollina Happy to learn here. What would you suggest instead of apply, pass the logger into the function?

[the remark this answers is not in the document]
markrzen: Or is `call` the more performant of the two for whatever reason?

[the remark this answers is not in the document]
mcollina: call is better, yes.

Also, I would just do a function call without setting `this`.

mcollina:
Before landing this we should check if this degrades performance. The load tests are in https://github.com/fastify/fastify-http-proxy/tree/master/benchmarks

markrzen:
Sorry for the slow down on this. Still looking into why the tests on master aren't running successfully on my machine.

markrzen:
Dropped the usage of this, and simply passing in the logger instance into the logging methods. Hopefully that is what you meant by doing a function call without setting this.

markrzen:
I am unfamiliar with typescript, so please confirm those changes are correct.

README.md
```diff
@@ -199,6 +199,18 @@ Setting this option will not verify if the http method allows for a body.
 Override the `'Content-Type'` header of the forwarded request, if we are
 already overriding the [`body`][body].
 
+#### `logProxyError(log, err)`
```
Eomm, line 202: I would use the `request` object instead the `req.log` only as first parameter.

Since the feature aims to customize the logger, users may have some decorators they want to print out or some special headers

Do you agree?

README.md
```diff
@@ -199,6 +199,18 @@ Setting this option will not verify if the http method allows for a body.
...
+
+Override the default logging of proxy errors, allowing the changing of format or addition of information. You may also disable by setting to `false`.
+
+#### `logProxyRequest(log, source, requestConfig)`
+
+Override the default logging of proxy requests, allowing the changing of format or addition of information. You may also disable by setting to `false`.
+
+#### `logProxyResponse(log, res)`
```
Eomm, line 210:
```suggestion
#### `logProxyResponse(log, proxyResponse)`
```

README.md
markrzen: Happy to make that change. And I agree to that use case. I will wait for additional feedback, so I can get one last commit before hopefully merging.

markrzen:
This has fallen off my plate and I won't likely get back to it soon. Closing for now.

PR #87, opened 2020-06-30
Handle ENOTFOUND error, set default error code to 500 rather than 200

justinsc:
Hi there.  Small change to add error handling for failed DNS lookup when using http 1.  There was an existing test for http 2 only.  Also changes the default error code when an unknown error occurs to 500 rather than 200.

### DIFF
--- a/index.js
+++ b/index.js
@@ -103,12 +103,12 @@ module.exports = fp(function from (fastify, opts, next) {
       if (err) {
         this.request.log.warn(err, 'response errored')
         if (!this.sent) {
-          if (err.code === 'ERR_HTTP2_STREAM_CANCEL') {
+          if (err.code === 'ERR_HTTP2_STREAM_CANCEL' || err.code === 'ENOTFOUND') {
             this.code(503).send(new Error('Service Unavailable'))
           } else if (err instanceof TimeoutError) {
             this.code(504).send(new Error('Gateway Timeout'))
           } else {
-            this.send(err)
+            this.code(500).send(err)
           }
         }
         return
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "msgpack5": "^4.2.1",
     "nock": "^13.0.0",
     "pre-commit": "^1.2.2",
-    "proxyquire": "^2.1.0",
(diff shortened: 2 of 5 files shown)


index.js
```diff
@@ -103,12 +103,12 @@ module.exports = fp(function from (fastify, opts, next) {
...
         this.request.log.warn(err, 'response errored')
         if (!this.sent) {
-          if (err.code === 'ERR_HTTP2_STREAM_CANCEL') {
+          if (err.code === 'ERR_HTTP2_STREAM_CANCEL' || err.code === 'ENOTFOUND') {
             this.code(503).send(new Error('Service Unavailable'))
           } else if (err instanceof TimeoutError) {
             this.code(504).send(new Error('Gateway Timeout'))
           } else {
-            this.send(err)
+            this.code(500).send(err)
```
Eomm, line 111: Could you add a test for this?

This snipped should set the code by the user and then this change would fix it:

```
  instance.get('/', (request, reply) => {
    reply.code(201)
    reply.from()
  })
```

justinsc: Hi @Eomm. This is really tricky to test without adding some dependency injection into fastify-reply-from to allow a mock 'http' to be passed in to throw an unexpected error.  There are only a few ways to get `http.request` to throw a real error.  One complicated way is to make an http server that drops the incoming connection before completely sending the response.  However, if I wrote that test, then we'd probably want to specifically handle that error (ECONNRESET) and return '502 Bad Gateway' rather than allowing it to hit the catch-all... and therefore the catch-all would still be untested.

Thoughts or ideas?  I'm happy to the go down the dependency injection route, but that does add baggage to the implementation..

mcollina: You can use `proxyquire`.

justinsc: Thanks for the tip @mcollina  -- `proxyquire` worked great.

justinsc:
Thanks @Eomm and @mcollina!  Any way to get this into the 2.x version?  I'm not seeing a branch, or even a tag for `2.1.0`

Eomm:
I have created the v2.x branch 👍

PR #91, opened 2020-07-29
Support for latest undici and better spec compliance

mcollina:
#### Checklist


index.js
```diff
@@ -92,7 +92,7 @@ module.exports = fp(function from (fastify, opts, next) {
       if (body) {
         throw new Error('Rewriting the body when doing a GET is not allowed')
       }
-      headers['content-length'] = 0
+      // headers['content-length'] = 0
```
jsumners, line 95: Why not remove it completely?

mcollina: removed.

PR #104, opened 2020-09-29
Document onResponse replacing reply body

### LINKED ISSUE #103: content-length has to be deleted for replyOptions.onResponse
## 🐛 Bug Report

When using `reply.send(replacementResult)` from `replyOptions.onResponse` the `content-length` header is prepopulated.  This prevents sending of a longer result (I'm not sure what would happen with a shorter result).

## To Reproduce

```js
import {createReadStream} from 'fs';
import fastify from 'fastify';
import fastifyHttpProxy from 'fastify-http-proxy';

async function main() {
  const daemon1 = fastify();
  daemon1.get('/test', (_, reply) => {
    reply.header('x-accel-redirect', '/redirect-path');
    reply.send('');
  });
  await daemon1.listen();

  const upstream = `http://localhost:${daemon1.server.address().port}`;
  const daemon2 = fastify({logger: false});
  daemon2.register(fastifyHttpProxy, {
    upstream,
    prefix: '/',
    replyOptions: {
      onResponse: (request, reply, res) => {
        const destination = res.headers['x-accel-redirect'] ?? '';
        if (destination) {
          // XXX need to uncomment the following or it will fail
          // reply.removeHeader('content-length');
          return reply.send(createReadStream('package.json'));
        }

        reply.send(res);
      }
    }
  });

  daemon2.listen(8080);
}


main().catch(error => {
  console.error(error);
  process.exit(1);
});
```

Start this script then run `curl -v http://localhost:8080/`.  With `reply.removeHeader('content-length')` commented out you will receive a 0 byte response.

## Expected behavior

Removing the `content-type` header before sending the replacement reply causes the expected result (`./package.json` is transmitted in the response).  I'm not sure if this should be automatic or if this need should be documented here or on `fastify-reply-from`.  It took quite a lot for me to troubleshoot, specifically when I use `reply.send('redir')` from daemon1 I saw the first 5 characters of the response, this is when I realized the problem.

## Your Environment

- *node version*: 14
- *fastify version*: 3.4.1
- *fastify-http-proxy*: 4.0.4
- *fastify-reply-from*: 3.3.0
- *os*: Linux

coreyfarrell:
Fixes #103

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,18 @@ Called when an http response is received from the source.
 The default behavior is `reply.send(res)`, which will be disabled if the
 option is specified.
 
+When replying with a body of a different length it is necessary to remove
+the `content-length` header.
+
+```js
+{
+  onResponse: (request, reply, res) => {
+    reply.removeHeader('content-length');
+    reply.send('New body of different length');
+  }
+}
+```
+
 #### `onError(reply, error)`
 
 Called when an http response is received with error from the source.


mcollina (approved):
lgtm

PR #120, opened 2020-12-02
compatible fastify-reply-from

donghc:
compatible fastify-reply-from：When using HTTP, the request type is content- Type:application/x-www-form-urlencoded , Buffer.byteLength (body) error: the "string" argument must be of type string or an instance of buffer or arraybuffer. Received [object: null prototype]


jsumners (commented):
This will require test coverage.

mcollina:
Closing for no activity, feel free to reopen.

Issue #118, opened 2020-12-01
request.from(...) to a fastify service with multipart/form-data leads to ECONNRESET


theliuk:
## 🐛 Bug Report

I configured two web services, both running on the top of the fastify ecosystem:

- `my-reverse-proxy`: uses `fastify-multipart` because some endpoints (not reported in the example) have to use it. If incoming requests respect some logic, these are forwarded to...
    
- `my-end-service`: handles a http `multipart/form-data` call carrying a file and some metadata

## To Reproduce
I prepared [this repo](https://github.com/theliuk/fastify-reply-from-bug-repr) and you can follow the instruction provided in the README.md file

See [this repo](https://github.com/theliuk/fastify-reply-from-bug-repr)

## Expected behavior

The request if forwarded by `fastify-reply-from` to the service which has to handle the call. 
This service can handle all the parts of the `multipart/form-data` correctly.

## Your Environment

    - node: erbium
    - fastify: v3.8.0
    - fastify-reply-from: v3.4.0
    - fastify-multipart: v3.3.1
    - docker-compose version 1.27.4, build 40524192

Eomm:
What is the error?
Do you have any stacktrace?

Anyway I think the issue could be hidden here:

```
fastify.addHook('onRequest', async(req, res) => {
  //Perform some logic to decide wheter proxy the request or not
  return res.from('/')
})
```

`from` returns the `res` instance that is a thenable, so fastify is waiting to send something and this plugin is sending without notify fastify itself.

theliuk:
> What is the error?
> Do you have any stacktrace? (EDIT: in which instance of the two)
> 
> Anyway I think the issue could be hidden here:
> 
> ```
> fastify.addHook('onRequest', async(req, res) => {
>   //Perform some logic to decide wheter proxy the request or not
>   return res.from('/')
> })
> ```
> 
> `from` returns the `res` instance that is a thenable, so fastify is waiting to send something and this plugin is sending without notify fastify itself.

Consider logs of container `my-reverse-proxy-container`.
Here the stacktrace:

```
Error: HTTP request timed out
at ClientRequest.<anonymous> (/usr/src/app/node_modules/fastify-reply-from/lib/request.js:88:19)
at Object.onceWrapper (events.js:420:28)
at ClientRequest.emit (events.js:314:20)
at Socket.emitRequestTimeout (_http_client.js:715:9)
at Object.onceWrapper (events.js:420:28)
[...]

PR #121, opened 2020-12-02
added test for multipart/form-data request with text and file

theliuk:
I tried to reproduce [this bug](https://github.com/fastify/fastify-reply-from/issues/118) by means of a test. 
"Unfortunately" this new test passes and I wasn't able to reproduce the bug.

#### Checklist


mcollina:
I think you might want to add `fastify-multipart` directly... I think this plugin is incompatible with it.

theliuk:
> I think you might want to add `fastify-multipart` directly... I think this plugin is incompatible with it.

Indeed: If the "proxying instance" (using `fastify-reply-from`) registers `fastify-multipart`, the "target instance" doesn't get multipart data.

mcollina:
The problem is that we do not have:

1. the original data on the wire
2. a way to transform the data attached to `req` back into a multipart format

The most idiomatic solution to this problem at the user level is to leverage encapsulation.

```js
const app = Fastify()

app.register(async function (app) {
  app.register(Multipart)
  app.post(...) // route that uses multipart
})

app.register(async function (app) {
  app.register(From)
  app.post(...) // route that proxy correctly
})
``` 

I think the best path forward is to provide a meaningful error message when the plugin is registered and add some documentation in the README. You can use the `app.hasRequestDecorator` as a check, or maybe use `app.hasContentTypeParser` check. This is probably a stop-gap solution that at least provide help to the users.

We could also add a way to let the user provide the serialization method of the body.

theliuk:
> The problem is that we do not have:
> 
> 1. the original data on the wire
> 2. a way to transform the data attached to `req` back into a multipart format
> 
> The most idiomatic solution to this problem at the user level is to leverage encapsulation.
> 
> ```js
> const app = Fastify()
> 
> app.register(async function (app) {
>   app.register(Multipart)
>   app.post(...) // route that uses multipart
> })
> 
> app.register(async function (app) {
>   app.register(From)
>   app.post(...) // route that proxy correctly
> })
> ```
> 
> I think the best path forward is to provide a meaningful error message when the plugin is registered and add some documentation in the README. You can use the `app.hasRequestDecorator` as a check, or maybe use `app.hasContentTypeParser` check. This is probably a stop-gap solution that at least provide help to the users.
> 
> We could also add a way to let the user provide the serialization method of the body.

Ok, I'll do it.
Shall I keep the test which documents the incompatibility with `fastify-multipart`?

mcollina:
I think so. I'd also add it in the README if anybody stumbles on this.

README.md
```diff
@@ -12,6 +12,10 @@ HTTP2 to HTTP is supported too.
 npm i fastify-reply-from
 ```
 
+## Compatibility with fastify-multipart
+`fastify-reply-from` and [`fastify-multipart`](https://github.com/fastify/fastify-multipart) should not be registered as sibling plugins nor shold be registered in plugins which have a parent-child relationship.<br> The two plugins are incompatible, in the sense that the behavior of `fastify-reply-from` might not be the expected one when the above-mentioned conditions are not respected.<br>
+However, the two plugins may be used within the same fastify instance, at the condition that they belong to disjoint branches of the fastify plugins hierarchy tree.
```
mcollina, line 17: It would be good to explain why they are incompatible, i.e. the original multipart request is consumed when parsed.

mcollina (commented):
Good work, just two nits

mcollina (approved):
lgtm

PR #135, opened 2021-02-05
refactor(index): use template literal over string concat

Fdawgs:
#### Checklist


mcollina (changes requested):
Why?  This is totally safe as it is right now.

Fdawgs:
> 
> 
> Why? This is totally safe as it is right now.

Fair point! Also realized this won't work in in environments earlier than ES6.

PR #141, opened 2021-02-10
Address breaking changes undici@2=>@3

penx:
Address breaking changes introduced in undici@3

https://github.com/nodejs/undici/commit/9df73a1409769d25dab8d4d9dbeb8e11f0260763

#### Checklist

- [x] run `npm run test` ~and `npm run benchmark`~
*missing script: benchmark

### DIFF
--- a/index.js
+++ b/index.js
@@ -106,7 +106,7 @@ module.exports = fp(function from (fastify, opts, next) {
         if (!this.sent) {
           if (err.code === 'ERR_HTTP2_STREAM_CANCEL' || err.code === 'ENOTFOUND') {
             onError(this, { error: new createError.ServiceUnavailable() })
-          } else if (err instanceof TimeoutError || err.code === 'UND_ERR_HEADERS_TIMEOUT' || err.code === 'UND_ERR_BODY_TIMEOUT') {
+          } else if (err instanceof TimeoutError || err.code === 'UND_ERR_HEADERS_TIMEOUT') {
             onError(this, { error: new createError.GatewayTimeout() })
           } else {
             onError(this, { error: createError(500, err) })
--- a/index.js
+++ b/index.js
@@ -106,7 +106,7 @@ module.exports = fp(function from (fastify, opts, next) {
         if (!this.sent) {
           if (err.code === 'ERR_HTTP2_STREAM_CANCEL' || err.code === 'ENOTFOUND') {
             onError(this, { error: new createError.ServiceUnavailable() })
-          } else if (err instanceof TimeoutError || err.code === 'UND_ERR_HEADERS_TIMEOUT') {
+          } else if (err instanceof TimeoutError || err.code === 'UND_ERR_HEADERS_TIMEOUT' || err.code === 'UND_ERR_BODY_TIMEOUT') {
             onError(this, { error: new createError.GatewayTimeout() })
           } else {
             onError(this, { error: createError(500, err) })
--- a/index.js
+++ b/index.js
@@ -106,7 +106,7 @@ module.exports = fp(function from (fastify, opts, next) {
         if (!this.sent) {
           if (err.code === 'ERR_HTTP2_STREAM_CANCEL' || err.code === 'ENOTFOUND') {
             onError(this, { error: new createError.ServiceUnavailable() })
-          } else if (err instanceof TimeoutError || err.code === 'UND_ERR_REQUEST_TIMEOUT') {
+          } else if (err instanceof TimeoutError || err.code === 'UND_ERR_HEADERS_TIMEOUT') {
             onError(this, { error: new createError.GatewayTimeout() })
           } else {
             onError(this, { error: createError(500, err) })
--- a/test/undici-timeout.js
+++ b/test/undici-timeout.js
@@ -31,7 +31,7 @@ async function main () {
   instance.register(From, {
     base: `http://localhost:${target.server.address().port}`,
     undici: {
-      requestTimeout: 100
+      headersTimeout: 100
     }
   })
 
--- a/lib/request.js
+++ b/lib/request.js
@@ -112,6 +112,9 @@ function buildRequest (opts) {
         return
       }
 
+      // using delete, otherwise it will render as an empty string
+      delete res.headers['transfer-encoding']
+
       done(null, { statusCode: res.statusCode, headers: res.headers, stream: res.body })
     })
   }
--- a/lib/request.js
+++ b/lib/request.js
@@ -98,7 +98,8 @@ function buildRequest (opts) {
       method: opts.method,
       headers: Object.assign({}, opts.headers),
       body: opts.body,
-      requestTimeout: undiciOpts.requestTimeout
+      headersTimeout: undiciOpts.headersTimeout,
+      bodyTimeout: undiciOpts.bodyTimeout
     }
 
     // remove forbidden headers
--- a/test/undici-timeout-body-partial.js
+++ b/test/undici-timeout-body-partial.js
@@ -1,0 +1,60 @@
+'use strict'
+
+const t = require('tap')
+const http = require('http')
+const Fastify = require('fastify')
+const From = require('..')
+const got = require('got')
+const FakeTimers = require('@sinonjs/fake-timers')
+
+const clock = FakeTimers.createClock()
+
+t.autoend(false)
+
+const target = http.createServer((req, res) => {
+  t.pass('request proxied')
+  req.on('data', () => undefined)
+  req.on('end', () => {
+    res.writeHead(200)
+    res.flushHeaders()
+    res.write('test')
+    clock.setTimeout(() => {
+      res.end()
+      t.end()
+    }, 1000)
+  })
+})
+
+async function main () {
+  await target.listen(0)
+
+  const instance = Fastify()
+  t.tearDown(instance.close.bind(instance))
+  t.tearDown(target.close.bind(target))
+
+  instance.register(From, {
+    base: `http://localhost:${target.address().port}`,
+    undici: {
+      bodyTimeout: 100
+    }
+  })
+
+  instance.get('/', (request, reply) => {
+    reply.from()
+  })
+
+  await instance.listen(0)
+
+  try {
+    await got.get(`http://localhost:${instance.server.address().port}/`, { retry: 0 })
+  } catch (err) {
+    t.equal(err.code, 'ECONNRESET')
+    t.equal(err.response.statusCode, 200)
+    clock.tick(1000)
+    return
+  }
+
+  t.fail()
+}
+
+main()
--- a/test/undici-timeout-body.js
+++ b/test/undici-timeout-body.js
@@ -1,0 +1,63 @@
+'use strict'
+
+const t = require('tap')
+const http = require('http')
+const Fastify = require('fastify')
+const From = require('..')
+const got = require('got')
+const FakeTimers = require('@sinonjs/fake-timers')
+
+const clock = FakeTimers.createClock()
+
+t.autoend(false)
+
+const target = http.createServer((req, res) => {
+  t.pass('request proxied')
+  req.on('data', () => undefined)
+  req.on('end', () => {
+    res.flushHeaders()
+    clock.setTimeout(() => {
+      res.end()
+      t.end()
+    }, 1000)
+  })
+})
+
+async function main () {
+  await target.listen(0)
+
+  const instance = Fastify()
+  t.tearDown(instance.close.bind(instance))
+  t.tearDown(target.close.bind(target))
+
+  instance.register(From, {
+    base: `http://localhost:${target.address().port}`,
+    undici: {
+      bodyTimeout: 100
+    }
+  })
+
+  instance.get('/', (request, reply) => {
+    reply.from()
+  })
+
+  await instance.listen(0)
+
+  try {
+    await got.get(`http://localhost:${instance.server.address().port}/`, { retry: 0 })
+  } catch (err) {
+    t.equal(err.response.statusCode, 500)
+    t.deepEqual(JSON.parse(err.response.body), {
+      statusCode: 500,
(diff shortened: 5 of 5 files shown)


penx:
Moved to draft as I think docs need to be updated

penx:
Reviewed docs and I don't see any reference to `undici. requestTimeout`.

Would this be considered a breaking change though? As any upstream users using `undici. requestTimeout` will need to update their code.

mcollina:
Yes, I will bump the major once this lands.

mcollina:
This is missing the bodyTimeout and its relative error.

penx:
I've added the bodyTimeout option and handled the error and pushed, but I think this needs test coverage.

test/undici-timeout.js
```diff
@@ -31,7 +31,7 @@ async function main () {
   instance.register(From, {
     base: `http://localhost:${target.server.address().port}`,
     undici: {
-      requestTimeout: 100
+      headersTimeout: 100
```
mcollina, line 34: could you add a copy of this test with `bodyTimeout`  set instead? Thanks

mcollina (changes requested):

test/undici-timeout.js
penx: I created a test server that sends the headers before the body as follows:

```js
const target = http.createServer((req, res) => {
  t.pass('request proxied')
  req.on('data', () => undefined)
  req.on('end', () => {
    res.flushHeaders();
    clock.setTimeout(() => {
      res.end(JSON.stringify({ something: 'else' }));
      t.end()
    }, 1000)
  })
})
```

however `await got.get(`http://localhost:${instance.server.address().port}/`, { retry: 0 })` now throws with 

> RequestError: Parse Error: Content-Length can't be present with Transfer-Encoding

Am I right in thinking that the headers are forwarded to the client as soon as they are received, and then when this runs:

https://github.com/penx/fastify-reply-from/blob/bf4793159d1864f36e366c366098e60eccee99bd/index.js#L179-L181

We get an error because the headers have already been sent?

mcollina: Can you include a full stacktrace?

penx:
```
RequestError: Parse Error: Content-Length can't be present with Transfer-Encoding
    at ClientRequest.<anonymous> (/Users/penx/Development/forks/fastify-reply-from/node_modules/got/dist/source/core/index.js:953:111)
    at Object.onceWrapper (node:events:485:26)
    at ClientRequest.emit (node:events:390:22)
    at ClientRequest.origin.emit (/Users/penx/Development/forks/fastify-reply-from/node_modules/@szmarczak/http-timer/dist/source/index.js:39:20)
    at Socket.socketOnData (node:_http_client:509:9)
    at Socket.emit (node:events:378:20)
    at addChunk (node:internal/streams/readable:313:12)
    at readableAddChunk (node:internal/streams/readable:288:9)
    at Socket.Readable.push (node:internal/streams/readable:227:10)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
    at Socket.socketOnData (node:_http_client:502:22)
    at Socket.emit (node:events:378:20)
    at addChunk (node:internal/streams/readable:313:12)
    at readableAddChunk (node:internal/streams/readable:288:9)
    at Socket.Readable.push (node:internal/streams/readable:227:10)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
    at TCP.callbackTrampoline (node:internal/async_hooks:132:14) {
  code: 'HPE_UNEXPECTED_CONTENT_LENGTH',
  timings: {
    start: 1612976101252,
    socket: 1612976101253,
    lookup: 1612976101254,
    connect: 1612976101255,
    secureConnect: undefined,
    upload: 1612976101255,
    response: undefined,
    end: undefined,
    error: 1612976103275,
    abort: undefined,
    phases: {
      wait: 1,
      dns: 1,
      tcp: 1,
      tls: undefined,
      request: 0,
      firstByte: undefined,
      download: undefined,
      total: 2023
    }
   }
 }
```

mcollina: I think you might need to remove `transfer-encoding` from `headers` in https://github.com/fastify/fastify-reply-from/blob/d4729809effa27d7429e41098938c0bd2dca4051/lib/request.js#L114.

penx: This helps, but then the `UND_ERR_BODY_TIMEOUT` can't be intercepted without an unwanted performance hit, as far as I can tell.

It seems to land here:
https://github.com/penx/fastify-reply-from/blob/bf4793159d1864f36e366c366098e60eccee99bd/index.js#L107-L107

with `err` being null, and then here:

https://github.com/penx/fastify-reply-from/blob/bf4793159d1864f36e366c366098e60eccee99bd/index.js#L136-L136

so starts streaming whatever it receives as the body from undici, in this case:

```json
{
   "statusCode": 500,                
   "code": "UND_ERR_BODY_TIMEOUT",   
   "error": "Internal Server Error", 
   "message": "Body Timeout Error",  
} 
```

The only approach to intercept this I guess would be to wait for the stream to complete first. Pretty sure this wouldn't be wanted so I guess we have the unit test expect this response?

mcollina: I'm pretty sure the body you linked is not what is get streamed out from undici. https://github.com/nodejs/undici/blob/master/lib/core/errors.js#L21 as you can see this is an actual error being emitted.

index.js
```diff
@@ -106,7 +106,7 @@ module.exports = fp(function from (fastify, opts, next) {
         if (!this.sent) {
           if (err.code === 'ERR_HTTP2_STREAM_CANCEL' || err.code === 'ENOTFOUND') {
             onError(this, { error: new createError.ServiceUnavailable() })
-          } else if (err instanceof TimeoutError || err.code === 'UND_ERR_REQUEST_TIMEOUT') {
+          } else if (err instanceof TimeoutError || err.code === 'UND_ERR_HEADERS_TIMEOUT' || err.code === 'UND_ERR_BODY_TIMEOUT') {
```
mcollina, line 109: Do you think `UND_ERR_BODY_TIMEOUT` could happen here? From what I've read it will happen after we have started streaming the response.

mcollina (changes requested):

index.js
penx: I think you're right, but I'm going to need to read some more of undici's source code to understand the flow here. I'll pick this up tomorrow morning.

penx: i.e.

https://github.com/nodejs/undici/blob/c4e237067dbc784d1cefa2b80a3ca5213a3561c8/lib/core/client.js#L731-L735

https://github.com/nodejs/undici/blob/c4e237067dbc784d1cefa2b80a3ca5213a3561c8/lib/core/client.js#L522-L524      ? setTimeout(onBodyTimeout, client[kBodyTimeout], this)
      : null

penx: I think more thought is needed here.

If a user sets `bodyTimeout` but not `headersTimeout`, what would they expect the behaviour to be? And what are our options on handling this?

There are a few edge cases that would need to be considered and perhaps tested for:


## Case 1

The headers take longer to respond than `bodyTimeout`, but the subsequent time taken to send the body being less than `bodyTimeout`.

This should complete successfully, the headers can take as long as undici allows them to, `bodyTimeout` is the time taken from the headers being sent to the body being closed (`res.end()`).

I can add a test for this.

## Case 2

The headers are sent quickly, but the body is not sent at all within the `bodyTimeout`

At the moment, we send response code 500 to the client, along with the error from Undici in the body.

https://github.com/penx/fastify-reply-from/blob/bf4793159d1864f36e366c366098e60eccee99bd/index.js#L130-L130

`stream.destroy(<BodyTimeoutError>)`
https://github.com/nodejs/undici/blob/7f3eef73355ced497107023a7af8b987fd01459c/lib/core/util.js#L105

```json
{
  statusCode: 500,
  code: 'UND_ERR_BODY_TIMEOUT',
  error: 'Internal Server Error',
  message: 'Body Timeout Error'
}
```

## Case 3

The headers being sent quickly, but the body being *partially sent* within the `bodyTimeout` period.

```js
const target = http.createServer((req, res) => {
  t.pass('request proxied')
  req.on('data', () => undefined)
  req.on('end', () => {
    res.flushHeaders()
    res.write('test')
    clock.setTimeout(() => {
      res.end()
      t.end()
    }, 1)
  })
})
```

This partially streams a response ("test") to the client and then hits node internals `stream.destroy(<BodyTimeoutError>)` (as in Case 2): 

https://github.com/nodejs/undici/blob/7f3eef73355ced497107023a7af8b987fd01459c/lib/core/util.js#L105

However on this occasion, node internals seem to behave differently because the body has already been partially streamed.

`got.get` throws a "The server aborted pending request" error, and reports that the response code is '200'. `got` does not allow us to access the partially streamed response, as far as I can see.

Confirmed looking at a local proxy that the response code remains as 200 and "test" is sent to the client.

---

I don't think the behaviour of 2 and 3 are great, but I'm going to need to do some more reading to figure out if we have any options to improve them.

penx: > Do you think UND_ERR_BODY_TIMEOUT could happen here? From what I've read it will happen after we have started streaming the response.

Perhaps if the response code changes to 500 before we start streaming the body to the client, we could intercept it 🤷 

But as the code stands at the moment, no I don't think `UND_ERR_BODY_TIMEOUT` would be intercepted.

mcollina: As for the HTTP spec, headers are alway sent _before_ the body. If they are _after_ the body are called trailers. I don't think this module supports trailers.

penx: does that include any response code change that is sent after `flushHeaders` but before anything is written to the body?

penx: ah I've been using `res.flushHeaders()` on the fake servers when I should probably be using `res.writeHead(200)` or similar, which changes things a bit

penx: ok so in case 2, the headers have not yet been sent to the client, so when an error is received from the target server, and we call `payload.pipe(res)`, something in node internals seems to change the statusCode from 200 to 500.

In case 3, the headers have already been sent to the client, and so node has no choice but to terminate the response.

We may be able to do something in case 2 but I'd need to read the node internals code to see what it's doing when it changes the status to 500.

mcollina: Read up on https://developer.mozilla.org/en-US/docs/Web/HTTP/Messages.

TL;DR the order is start lines (status code, etc), headers, body, trailers.

> In case 3, the headers have already been sent to the client, and so node has no choice but to terminate the response (well I guess it could send a trailer as you say, but it doesn't seem to).

Yes. The response has already been sent. You'll get an error on the other side.

penx: ok looks like I can improve case 2 by extracting the error handler and setting it when before calling `this.send(res.stream)`

i.e.

```js
 const errorHandler = (err, request, reply) => {
      reply.request.log.warn(err, 'response errored')
        if (!reply.sent) {
          if (err.code === 'ERR_HTTP2_STREAM_CANCEL' || err.code === 'ENOTFOUND') {
            onError(reply, { error: new createError.ServiceUnavailable() })
          } else if (err instanceof TimeoutError || err.code === 'UND_ERR_HEADERS_TIMEOUT' || err.code === 'UND_ERR_BODY_TIMEOUT') {
            onError(reply, { error: new createError.GatewayTimeout() })
          } else {
            onError(reply, { error: createError(500, err) })
          }
        }
    }

    request({ method: req.method, url, qs, headers: requestHeaders, body }, (err, res) => {
      if (err) {
        errorHandler(err, this.request, this)
        return
      }
      this.request.log.info('response received')
      if (sourceHttp2) {
        copyHeaders(
          rewriteHeaders(stripHttp1ConnectionHeaders(res.headers)),
          this
        )
      } else {
        copyHeaders(rewriteHeaders(res.headers), this)
      }
      this.code(res.statusCode)
      if (onResponse) {
        onResponse(this.request.raw, this, res.stream)
      } else {
        this.context.errorHandler = errorHandler
        this.send(res.stream)
      }
    })
    return this
  })
```

penx: (note `errorHandler` doesn't fire on case 3 and I'm not sure there's much we could do other than terminating the response, I can add a test case for this)

index.js
```diff
@@ -127,6 +131,7 @@ module.exports = fp(function from (fastify, opts, next) {
       if (onResponse) {
         onResponse(this.request.raw, this, res.stream)
       } else {
+        this.context.errorHandler = errorHandler
```
mcollina, line 134: you can't do this :(, it will leak memory!

mcollina (changes requested):

index.js
penx: Thanks, I'd meant to check on the best way to do this. I'm hoping this is the correct way:

https://github.com/fastify/fastify-reply-from/pull/141/commits/45bbd9da38389be8a84ff37f0c0df77902d08c6c

index.js
```diff
@@ -29,6 +29,9 @@ module.exports = fp(function from (fastify, opts, next) {
     sessionTimeout: opts.sessionTimeout,
     undici: opts.undici
   })
+
+  fastify.setErrorHandler((error, request, reply) => errorHandler(error, request, reply, onErrorDefault))
```
mcollina, line 33: unfortunately this is not viable either. This will be implemented by the developers using the plugin.

penx: are you saying that `fastify.setErrorHandler` may be used higher up in the context and this would override it?

Perhaps there is a way we can intercept and change the error but still pass it up to the parent context error handler?

mcollina: I don't think we should intercept that error.

mcollina: adding an error handler here will override any other that were previously set.

penx: It would be nice for the user to get a 504 rather than 500, but I agree it may be overkill to handle this at the moment.

I've reverted the last few commits, removed the redundant check for UND_ERR_BODY_TIMEOUT and added a test.

mcollina (approved):
lgtm

PR #143, opened 2021-02-12
Use undici by default

mcollina:
#### Checklist

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -7,6 +7,7 @@ const eos = require('end-of-stream')
 const pump = require('pump')
 const undici = require('undici')
 const { stripHttp1ConnectionHeaders } = require('./utils')
+const http2 = require('http2')
 
 class TimeoutError extends Error {}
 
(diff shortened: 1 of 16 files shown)


lib/request.js
```diff
@@ -102,6 +105,12 @@ function buildRequest (opts) {
       bodyTimeout: undiciOpts.bodyTimeout
     }
 
+    if (opts.url.protocol.indexOf('unix') === 0) {
+      done(new Error('unix socket not supported with undici yet'))
```
jsumners, line 109: This seems like a major obstacle to using it as the default. Would it not be better to solve this before switching this default?

fox1t: Maybe we can use undici as default and add a section in docs that points out that if `unix:` is needed `opts.undici` has to be set to false?

fox1t (approved):

kibertoad:
Should we maybe bump node requirement from 8 to 10, since neither undici nor this plugin are tested on Node 8, so there is no guarantee either is still working?..

lib/request.js
mcollina: I have added back most of the unix socket support I could add. I do not think supporting unix sockets as in `reply.from(path)` is useful. It will be supported as a base.

mcollina:
@jsumners PTAL

jsumners (approved):
LGTM

kibertoad (approved):

PR #157, opened 2021-03-21
158: Added a hook to get the url base based on URL object

yohay-ma:
see: https://github.com/fastify/fastify-reply-from/issues/158
Needed for a gradual rollout of services.

For example: 
instance: http://localhost:3000/test
service A: http://localhost:3001/test
service B: http://localhost:3002/test

I'd like to set 20% of the requests to go to http://localhost:3001/test and 80% to go to http://localhost:3001/test

In this case the getUpstream will do a random of 20%/80% between the two

#### Checklist

- [v ] run `npm run test` and `npm run benchmark`
- [v ] tests and/or benchmarks are included
- [ v] documentation is changed or added
- [ v] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)


mcollina (changes requested):
Good work! This would be useful. This PR adds another option called `disableCache`. Can you document it? 

The docs should note that generating the URLs is one of the main bottleneck of this module, so disabling this will effect performance negatively.

yohay-ma:
> Good work! This would be useful. This PR adds another option called `disableCache`. Can you document it?
> 
> The docs should note that generating the URLs is one of the main bottleneck of this module, so disabling this will effect performance negatively.

Added disableCache to readme and ts interface

mcollina (approved):
lgtm

Issue #117, opened 2020-11-30
The \"string\" argument must be of type string or an instance of Buffer or ArrayBuffer. Received [Object: null prototype]


donghc:
## 🐛 Bug Report

When using HTTP, the request type is content- Type:application/x-www-form-urlencoded , Buffer.byteLength (body) error: the "string" argument must be of type string or an instance of buffer or arraybuffer. Received [object: null prototype]

## To Reproduce

```
curl -X POST "http://127.0.0.1:3000/changeSignStatus" -H "accept:*/*" -H "Content-Type:application/x-www-form-urlencoded; charset=UTF-8" -d "key=1152&status=1"
```

## Your Environment

- *node version*: v14.4.0
- *fastify version*: ^3.8.0
- *os*: Mac
- *any other relevant information*

mcollina:
Can you provide an example on how to reproduce the problem? Note that you need a content-type parser to pass this through successfully - see fastify-http-proxy.

donghc:
I'm sorry for the delay. This is an error report. This is another plug-in problem.

donghc:
> 您能否提供一个有关如何重现该问题的示例？请注意，您需要一个内容类型解析器才能成功通过此解析器-请参见fastify-http-proxy。
for example
```
const Fastify = require('fastify')
const server = Fastify({logger:true})
const reply_form = require('fastify-reply-from')

server.register(require('fastify-formbody'))
server.register(reply_form)
server.register(require('fastify-http-proxy'), {
    upstream: 'http://division-test.haojile.net.cn/',
    prefix: '/division', // optional
    rewritePrefix: '/api/division',
    http2: false // optional
})

server.listen(3000)
```
```
curl -X POST "http://127.0.0.1:3000/division/global/changeSignStatus" -H "accept:*/*" -H "Content-Type:application/x-www-form-urlencoded; charset=UTF-8" -d "key=1152&status=1"
```

donghc:
I find index.js Line 79 in the file，        
```
body = shouldEncodeJSON ? JSON.stringify(this.request.body) :this.request.body 
headers['content-length'] = Buffer.byteLength(body)
```
body is object，so throw an exception。

mcollina:
There is some conflict with fastify-formbody, remove that and fastify-http-proxy would work.

mcollina:
This needs a fix anyway, would you like to send a PR?

donghc:
> This needs a fix anyway, would you like to send a PR?

[...]

PR #160, opened 2021-03-30
fix: encoding when used with fastify-formbody plugin

skywickenden:
This fixes https://github.com/fastify/fastify-reply-from/issues/117 which is caused by combining `fastify-reply-from` with `fastify-formbody` 

#### Checklist

- [x] run `npm run test` and `npm run benchmark` (no benchmark on this plugin)

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,10 +32,6 @@ module.exports = fp(function from (fastify, opts, next) {
     const rewriteRequestHeaders = opts.rewriteRequestHeaders || requestHeadersNoOp
     const getUpstream = opts.getUpstream || upstreamNoOp
     const onError = opts.onError || onErrorDefault
-    const contentTypesToEncode = new Set([
-      'application/json',
-      ...(opts.contentTypesToEncode || [])
-    ])
 
     if (!source) {
       source = req.url
@@ -80,6 +76,10 @@ module.exports = fp(function from (fastify, opts, next) {
         // detect if body should be encoded as JSON
         // supporting extended content-type header formats:
         // - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type
+        const contentTypesToEncode = new Set([
+          'application/json',
+          ...(opts.contentTypesToEncode || [])
+        ])
         const lowerCaseContentType = contentType.toLowerCase()
         const plainContentType = lowerCaseContentType.indexOf(';') > -1
           ? lowerCaseContentType.slice(0, lowerCaseContentType.indexOf(';'))
--- a/index.js
+++ b/index.js
@@ -32,6 +32,10 @@ module.exports = fp(function from (fastify, opts, next) {
     const rewriteRequestHeaders = opts.rewriteRequestHeaders || requestHeadersNoOp
     const getUpstream = opts.getUpstream || upstreamNoOp
     const onError = opts.onError || onErrorDefault
+    const contentTypesToEncode = new Set([
+      'application/json',
+      ...(opts.contentTypesToEncode || [])
+    ])
 
     if (!source) {
       source = req.url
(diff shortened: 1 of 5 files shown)


test/post-formbody.js
```diff
@@ -0,0 +1,57 @@
...
+const get = require('simple-get').concat
+
+const instance = Fastify()
+instance.register(From)
+instance.register(require('fastify-formbody'))
+
+t.plan(9)
+t.tearDown(instance.close.bind(instance))
+
+const target = http.createServer((req, res) => {
```
simoneb, line 16: props for the thorough testing

index.js
```diff
@@ -76,7 +76,13 @@ module.exports = fp(function from (fastify, opts, next) {
         // detect if body should be encoded as JSON
         // supporting extended content-type header formats:
         // - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type
-        const shouldEncodeJSON = contentType.toLowerCase().indexOf('application/json') === 0
+        const contentTypesNotToEncode = [
+          'application/json',
+          'application/x-www-form-urlencoded'
+        ]
```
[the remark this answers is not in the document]
skywickenden: Do you want `'application/x-www-form-urlencoded'` in as a default or should I document the need to pass it in when working with `fastify-formbody`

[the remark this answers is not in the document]
jsumners: I think the doc recommendation would be best. We don't know how people will want to handle the form mimetype. The json mimetype is straight forward.

index.js
```diff
@@ -32,6 +32,10 @@ module.exports = fp(function from (fastify, opts, next) {
     const rewriteRequestHeaders = opts.rewriteRequestHeaders || requestHeadersNoOp
     const getUpstream = opts.getUpstream || upstreamNoOp
     const onError = opts.onError || onErrorDefault
+    const contentTypesToEncode = new Set([
+      'application/json',
+      ...(opts.contentTypesToEncode || [])
+    ])
```
mcollina, line 38: This set and the relative option should be moved one level up as creating a new set every time is relatively expensive.

index.js
```diff
@@ -76,7 +80,8 @@ module.exports = fp(function from (fastify, opts, next) {
         // detect if body should be encoded as JSON
         // supporting extended content-type header formats:
         // - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type
-        const shouldEncodeJSON = contentType.toLowerCase().indexOf('application/json') === 0
+        const plainContentType = contentType.toLowerCase().split(';')[0]
```
mcollina, line 83: Please avoid the use of split() as it allocates a new array. You can use `str.slice(0, str.indexOf(':'))` instead

index.js
```diff
@@ -76,7 +76,15 @@ module.exports = fp(function from (fastify, opts, next) {
...
-        const shouldEncodeJSON = contentType.toLowerCase().indexOf('application/json') === 0
+        const contentTypesToEncode = new Set([
+          'application/json',
+          ...(opts.contentTypesToEncode || [])
+        ])
+        const lowerCaseContentType = contentType.toLowerCase()
+        const plainContentType = lowerCaseContentType.indexOf(';') > -1
+          ? lowerCaseContentType.slice(0, lowerCaseContentType.indexOf(';'))
+          : lowerCaseContentType
+        const shouldEncodeJSON = contentTypesToEncode.has(plainContentType)
```
mcollina, line 87: I meant that this needs to go on like 19 and it should be an option of the plugin.

skywickenden:
@mcollina I don't know why on-error test is timing out on 10macOS. It runs fine almost instantly when I run it locally (using mac and node 10) I also can't see why that change would cause this error. I can't see any option to rerun the action. Any ideas?

simoneb:
> @mcollina I don't know why on-error test is timing out on 10macOS. It runs fine almost instantly when I run it locally (using mac and node 10) I also can't see why that change would cause this error. I can't see any option to rerun the action. Any ideas?

try force pushing an empty commit to trigger another build

index.js
skywickenden: The code that you selected above cannot be moved up as it depends on the current request for the content type. I think  you meant for me to move the `contentTypesToEncode` option onto the register options rather than the `decorateReply`, which I have now done.

mcollina: This does what I meant.

jsumners (approved):
LGTM

mcollina (approved):
lgtm

PR #167, opened 2021-04-23
Retry connection on ECONNRESET

artur-ma:
Trying to address https://github.com/fastify/help/issues/413 , still in the draft since I want to understand if I'm in the right direction, if so I will continue with the PR and will add tests

#### Checklist


index.js
```diff
@@ -28,6 +28,34 @@ module.exports = fp(function from (fastify, opts, next) {
     base,
     undici: opts.undici
   })
+
+  function createRetryRequest (requestImpl, reply, retriesCount) {
+    function requestRetry (req, cb) {
+      let retries = 0
+
+      function run () {
+        requestImpl(req, function (err, res) {
```
mcollina, line 37: You can't retry POSTs

mcollina (changes requested):

mcollina:
(Good work anyway!)

index.js
artur-ma: Retry methods restricted

artur-ma:
I have concerns regarding `createRequestRetry` I'm creating new wrapper on each request, which may affect performance, but I'm not sure how to avoid it

mcollina:
We would need to benchmark it.

artur-ma:
Added tests, I will run benchmark and post results to see if there is a huge penalty 

btw, undici will not throw `ECONNRESET` it has its own error, code, should I support it too? WDYT?

mcollina:
Yes please. I do not recall what code undici throws!

index.js
```diff
@@ -20,14 +20,20 @@ module.exports = fp(function from (fastify, opts, next) {
     'application/json',
     ...(opts.contentTypesToEncode || [])
   ])
+
+  const retryMethods = new Set(opts.retryMethods || [
+    'GET', 'HEAD', 'PUT', 'DELETE', 'OPTIONS', 'TRACE'
```
mcollina, line 25: PUT and DELETE should not be retriable.

mcollina (commented):
Good work! This needs docs.

artur-ma:
without retry wrapper

image

with the wrapper

image


looks like no diff at all

index.js
artur-ma: will remove them, although MDN says they are indepotent

mcollina (approved):
lgtm

README.md
```diff
@@ -189,6 +189,13 @@ This only applies when a custom [`body`](#body) is not passed in. Defaults to:
 ]
 ```
 
+#### `retryMethods`
+
+On which methods should the connection be retried in case of socket hang up.  
+**Be aware** that setting here not idempotent method may lead to unexpected results on target.
+
+By default: `GET`, `HEAD`, `OPTIONS`, `TRACE`
```
Fdawgs, line 197: Can't seem to put this inside a suggestion. 😞 
I would suggest changing this to something like the following, to indicate it accepts an array of options, as at present it the documentation makes it looks like it accepts a single string:

Defaults to:
```js
[ 'GET', 'HEAD',  'OPTIONS', 'TRACE' ]
```

artur-ma: Done

mcollina (approved):
lgtm

PR #179, opened 2021-06-14
Upgrade undici and update deps

simone-sanfratello:
Upgrade undici to `v4`

Note the proposal of drop the node `v10` support because `undici` is not supporting it

~~typecheck fails due to https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53820, so I'm also waiting for that~~

#### Checklist


simone-sanfratello:
on node `v10` is failing because of `queueMicrotask`

https://github.com/nodejs/undici/blob/main/lib/api/api-request.js#L135

I see `undici` is not supporting node `v10` so I guess the consistent solution is to drop node `v10` on `fastify-reply-from` too

kibertoad:
@simone-sanfratello `fastify` itself supports down to Node 10.16.0, so I don't think that we can remove support for 10.x until the next fastify semver major.

simone-sanfratello:
we have 2 options, either add node `v10` support on `undici` or drop node `v10` in `fastify-reply-from`

I'm proposing this PR for the second solution

kibertoad:
@simone-sanfratello What I'm trying to say, it's going to be confusing for users if Node requirements for one of the plugins are stricter than the requirements for fastify itself, and it will block all future updates to the plugin for everyone on older Node versions. This may prevent this PR from landing in the nearby future.

mcollina:
+1 to drop support for Node 10.

kibertoad:
@mcollina Should there be an explicit text in readme stating that only Node 12+ is supported and users of older node should use older version of the plugin, ir just the engine entry is sufficient?

package.json
```diff
@@ -27,39 +27,39 @@
...
+    "end-of-stream": "^1.4.4",
     "fastify-plugin": "^3.0.0",
     "http-errors": "^1.8.0",
     "pump": "^3.0.0",
-    "semver": "^7.2.1",
-    "tiny-lru": "^7.0.0",
-    "undici": "^3.3.1"
+    "semver": "^7.3.5",
+    "tiny-lru": "^7.0.6",
+    "undici": "^4.0.0-rc.7"
```
kibertoad, line 62: 4.0.0 has already landed

simone-sanfratello:
> @mcollina Should there be an explicit text in readme stating that only Node 12+ is supported and users of older node should use older version of the plugin, or just the engine entry is sufficient?

imo this should be a major release, but happy yo add some lines in the readme to make it clear

mcollina:
Yes, it's a semver-major change.

lib/request.js
```diff
@@ -120,7 +120,9 @@ function buildRequest (opts) {
       done(new Error('unix socket not supported with undici yet'))
       return
     } else {
-      pool = undiciAgent.get(baseUrl || opts.url.origin)
+      const options = getUndiciOptions(opts.undici)
+      options.dispatcher = undiciAgent
+      pool = new undici.Client(baseUrl || opts.url.origin, options)
```
mcollina, line 125: This should be a Pool, not a client.

mcollina (changes requested):

lib/request.js
```diff
@@ -120,7 +120,9 @@ function buildRequest (opts) {
       done(new Error('unix socket not supported with undici yet'))
       return
     } else {
-      pool = undiciAgent.get(baseUrl || opts.url.origin)
+      const options = getUndiciOptions(opts.undici)
+      options.dispatcher = undiciAgent
```
mcollina, line 124: are you sure this is needed?

simone-sanfratello: I think so, because of the previous instruction `pool = undiciAgent.get`

mcollina (approved):
lgtm

Fdawgs (approved):
👌🏻 Good job!

Issue #193, opened 2021-09-20
Handle 503 and Retry-After header


mcollina:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

After receiving a 503, we might want to retry sending that request after a short delay.

This should be done for GET requests.

leorossi:
I am working on this.
Should _all_ GET requests be retried? What happens when specifically configured not to retry?
What if `Retry-After` header is not set?

mcollina:
> Should all GET requests be retried? 

Ideally yes

> What happens when specifically configured not to retry?

How is this configuration happen? Ideally we should support a flag to avoid retries and set a maximum time to retry.

mcollina:
If there is no Retry-After, just retry after `42` ms and then use exponential backoff.

leorossi:
From test:

`instance.register(From, { undici: true, retryMethods: ['DELETE'] })`

In this case, should I still retry on GET and a 503 response?

mcollina:
I don't understand the question.

leorossi:
So, you can specify to the plugin which methods needs to be _retried_ . 
Do we want to retry always the `GET` methods even though the user specified _not_ to retry it?

mcollina:
nope, the users win

PR #200, opened 2021-09-29
Feature/retry on 503

leorossi:
Will _always_ retry on 503 errors, unless the user does not include `GET` in `retryMethods` option.

Fixes: #193 

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,16 +42,12 @@ module.exports = fp(function from (fastify, opts, next) {
     const rewriteRequestHeaders = opts.rewriteRequestHeaders || requestHeadersNoOp
     const getUpstream = opts.getUpstream || upstreamNoOp
     const onError = opts.onError || onErrorDefault
-    let retriesCount = opts.retriesCount || 0
+    const retriesCount = opts.retriesCount || 0
 
     if (!source) {
       source = req.url
     }
 
-    if (retryMethods.has('GET') && req.method === 'GET' && retriesCount < 1) {
-      // force retries on GET requests
-      retriesCount = 1
-    }
     // we leverage caching to avoid parsing the destination URL
     const dest = getUpstream(req, base)
     let url
@@ -122,7 +118,7 @@ module.exports = fp(function from (fastify, opts, next) {
     const requestHeaders = rewriteRequestHeaders(req, headers)
     const contentLength = requestHeaders['content-length']
     let requestImpl
-    if (retriesCount && retryMethods.has(req.method) && !contentLength) {
+    if (retryMethods.has(req.method) && !contentLength) {
       requestImpl = createRequestRetry(request, this, retriesCount, retryOnError, 503)
     } else {
       requestImpl = request
@@ -222,20 +218,26 @@ function createRequestRetry (requestImpl, reply, retriesCount, retryOnError, ret
 
     function run () {
       requestImpl(req, function (err, res) {
-        let retryAfter = 42
+        let retryAfter = 42 * Math.random() * (retries + 1)
         if (res && res.headers['retry-after']) {
           retryAfter = res.headers['retry-after']
         }
-        if (!reply.sent && retriesCount > retries) {
-          if ((err && err.code === retryOnError) ||
-          (res && res.statusCode === retryOnCode && req.method === 'GET')) {
-            retries += 1
-            setTimeout(run, retryAfter)
-            return
+
+        if (!reply.sent) {
+          // always retry on 503 errors
+          if (res && res.statusCode === retryOnCode && req.method === 'GET') {
+            return retry(retryAfter)
+          } else if (retriesCount > retries && err && err.code === retryOnError) {
+            return retry(retryAfter)
           }
         }
         cb(err, res)
       })
+    }
+
+    function retry (after) {
+      retries += 1
+      setTimeout(run, after)
     }
 
     run()
--- a/index.js
+++ b/index.js
@@ -42,12 +42,16 @@ module.exports = fp(function from (fastify, opts, next) {
     const rewriteRequestHeaders = opts.rewriteRequestHeaders || requestHeadersNoOp
     const getUpstream = opts.getUpstream || upstreamNoOp
     const onError = opts.onError || onErrorDefault
-    const retriesCount = opts.retriesCount || 0
+    let retriesCount = opts.retriesCount || 0
 
     if (!source) {
       source = req.url
     }
 
+    if (retryMethods.has('GET') && req.method === 'GET' && retriesCount < 1) {
+      // force retries on GET requests
+      retriesCount = 1
+    }
     // we leverage caching to avoid parsing the destination URL
     const dest = getUpstream(req, base)
     let url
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -213,21 +216,24 @@ function isFastifyMultipartRegistered (fastify) {
...
-function createRequestRetry (requestImpl, reply, retriesCount, retryOnError) {
+function createRequestRetry (requestImpl, reply, retriesCount, retryOnError, retryOnCode = 503) {
   function requestRetry (req, cb) {
     let retries = 0
 
     function run () {
       requestImpl(req, function (err, res) {
-        if (err && !reply.sent && retriesCount > retries) {
-          if (err.code === retryOnError) {
+        let retryAfter = 42
```
mcollina, line 225: Please add a comment for this.

mcollina:
```suggestion
        // Magic number, so why not 42? We might want to make this configurable.
        let retryAfter = 42
```

index.js
```diff
@@ -213,21 +216,25 @@ function isFastifyMultipartRegistered (fastify) {
...
+function createRequestRetry (requestImpl, reply, retriesCount, retryOnError, retryOnCode = 503) {
   function requestRetry (req, cb) {
     let retries = 0
 
     function run () {
       requestImpl(req, function (err, res) {
-        if (err && !reply.sent && retriesCount > retries) {
-          if (err.code === retryOnError) {
+        // Magic number, so why not 42? We might want to make this configurable.
+        let retryAfter = 42
```
mcollina, line 226:
```suggestion
        let retryAfter = 42 * Math.random() * (retries + 1)
```

This is probably better as it is not fixed.

index.js
```diff
@@ -213,25 +212,36 @@ function isFastifyMultipartRegistered (fastify) {
   return fastify.hasContentTypeParser('multipart') && fastify.hasRequestDecorator('multipart')
 }
 
-function createRequestRetry (requestImpl, reply, retriesCount, retryOnError) {
+function createRequestRetry (requestImpl, reply, retriesCount, retryOnError, retryOnCode = 503) {
```
Eomm, line 215:
```suggestion
function createRequestRetry (requestImpl, reply, retriesCount, retryOnError, retryOnCode) {
```

test/retry-on-503.js
```diff
@@ -0,0 +1,80 @@
...
+      res.statusCode = 503
+      res.setHeader('Content-Type', 'text/plain')
+      if (withRetryAfterHeader) {
+        res.setHeader('Retry-After', 1000)
+      }
+      return res.end('This Service Unavailable')
+    }
+    res.statusCode = 205
+    res.setHeader('Content-Type', 'text/plain')
+    return res.end('Hello World Twice!')
```
Eomm, line 24: I would improve the tests adding a check on the request count that has been done
```suggestion
    return res.end('Hello World: '+ requestCount)
```

test/retry-on-503.js
```diff
@@ -0,0 +1,80 @@
+'use strict'
+
+const { test } = require('tap')
+const Fastify = require('fastify')
+const From = require('..')
+const http = require('http')
+const got = require('got')
+
+let requestCount = 0
```
Eomm, line 9: Cannot be move in the `createTargetServer` function?

index.js
```diff
@@ -213,25 +212,36 @@ function isFastifyMultipartRegistered (fastify) {
...
-            run()
-            return
-          }
+        if (res && res.headers['retry-after']) {
+          retryAfter = res.headers['retry-after']
         }
 
+        if (!reply.sent) {
+          // always retry on 503 errors
+          if (res && res.statusCode === retryOnCode && req.method === 'GET') {
```
Eomm, line 230: if a server is returning 503 forever we will be stuck here. Maybe should check the retriesCount here also?

```suggestion
          if ((res && res.statusCode === retryOnCode && req.method === 'GET') || (retriesCount > retries && err && err.code === retryOnError)) {
```

index.js
leorossi: After some thoughts I will remove that, the issue is specifically for 503, and I would not make it configurable by the user since we may hit an external server.

index.js
leorossi: Good point! 

I updated the code and made a test for this case... although it doesn't work as expected and does not return, I cannot see why. 
@mcollina @Eomm need another pair of eyes [here](https://github.com/fastify/fastify-reply-from/blob/5628d50a52a05980b277a44f63d45e09c1141260/index.js#L217)

Eomm: It could be a timeout issue
The 10 retries take ~5-10 seconds, so I would reduce the `res.setHeader('Retry-After', 500)` or increase the `tap` timeout

leorossi: I already put a lower timeout with no success

Basically, this plugin will try 10 times (hardwired value) before moving on, if the target server returns 503 for `<= 10` times, it works... if it returns 503 for `> 10` it doesn't

climba03003: The problem is `got` hold the session forever.

test/retry-on-503.js
```diff
@@ -0,0 +1,102 @@
...
+  })
+
+  instance.get('/', (request, reply) => {
+    reply.from()
+  })
+
+  t.teardown(instance.close.bind(instance))
+  await instance.listen(0)
+
+  const res = await got.get(`http://localhost:${instance.server.address().port}`)
```
climba03003, line 97: we do not want got to retry.
```suggestion
  try {
    await got.get(`http://localhost:${instance.server.address().port}`, { retry: 0 })
    t.fail()
  } catch(err) {
    const res = err.response
    t.equal(res.headers['content-type'], 'text/plain')
    t.equal(res.statusCode, 503)
    t.equal(res.body.toString(), 'This Service is Unavailable')
  }
```

test/retry-on-503.js
```diff
@@ -0,0 +1,102 @@
...
+  t.equal(res.headers['content-type'], 'text/plain')
+  t.equal(res.statusCode, 205)
+  t.equal(res.body.toString(), 'Hello World 2!')
+  t.pass()
```
climba03003, line 75: `t.pass` is actual violating the test case. Remember to update the `t.plan`
```suggestion
```

test/retry-on-503.js
```diff
@@ -0,0 +1,102 @@
...
```
climba03003, line 49: `t.pass` is actual violating the test case. Remember to update the `t.plan`
```suggestion
```

climba03003 (changes requested):

test/retry-on-503.js
leorossi: I was missing that 🤦‍♂️

index.js
leorossi: Fixed!

climba03003 (approved):
LGTM.

index.js
```diff
@@ -215,23 +214,37 @@ function isFastifyMultipartRegistered (fastify) {
 
 function createRequestRetry (requestImpl, reply, retriesCount, retryOnError) {
   function requestRetry (req, cb) {
+    const MAX_RETRIES_ON_503 = 10
```
mcollina, line 217: can you make this configurable? Thanks

leorossi: done!

mcollina (approved):
lgtm

PR #237, opened 2022-03-15
Allow empty body override

jschumacher-dexcom:
Currently it is not possible to modify the proxied body to be undefined or an empty string since both of these values are falsy and the current implementation checks for the truthiness of `opts.body`.  This PR changes the behavior to check for the existence of `opts.body` instead (`"body" in opts`) to allow falsy values to be sent.

This would potentially be a breaking change since `fastify-reply-from` currently ignores the `body` param when it is set to `undefined` or an empty string.  The current empty string behavior feels unexpected, while the current behavior for `undefined` may be desired.  If that's the case, this could be reworked to have a more explicit way to strip the body, such as a new `doNotSendBody` option.

#### Test Results

```
  🌈 SUMMARY RESULTS 🌈


Suites:   72 passed, 72 of 72 completed
Asserts:  607 passed, of 607
Time:   20s
------------------------|---------|----------|---------|---------|-------------------
File                    | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
------------------------|---------|----------|---------|---------|-------------------
All files               |   97.98 |     95.7 |     100 |   98.64 |
 fastify-reply-from     |     100 |    99.05 |     100 |     100 |
  index.js              |     100 |    99.05 |     100 |     100 | 93
 fastify-reply-from/lib |   96.51 |    92.91 |     100 |   97.63 |
  request.js            |   96.47 |     92.3 |     100 |   97.84 | 47,51,224
  utils.js              |   96.66 |    95.65 |     100 |   96.66 | 9
------------------------|---------|----------|---------|---------|-------------------
```

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,7 @@ Replaces the original request body with what is specified. Unless
 [`contentType`](#contentType) is specified, the content will be passed
 through `JSON.stringify()`.
 Setting this option for GET, HEAD requests will throw an error "Rewriting the body when doing a {GET|HEAD} is not allowed".
+Setting this option to `undefined` will strip the body (and `content-type` header) entirely from the proxied request.
 
 #### `retriesCount`
 
--- a/README.md
+++ b/README.md
@@ -286,7 +286,7 @@ Replaces the original request body with what is specified. Unless
 [`contentType`](#contentType) is specified, the content will be passed
 through `JSON.stringify()`.
 Setting this option for GET, HEAD requests will throw an error "Rewriting the body when doing a {GET|HEAD} is not allowed".
-Setting this option to `undefined` will strip the body (and `content-type` header) entirely from the proxied request.
+Setting this option to `null` will strip the body (and `content-type` header) entirely from the proxied request.
 
 #### `retriesCount`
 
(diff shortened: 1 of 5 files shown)


mcollina:
The behavior for `undefined` is desired, but setting it to `null` or `''` should correctly be mapped to an empty body. 
Thanks for the work!

jschumacher-dexcom:
I've updated the behavior so that `undefined` will continue to be equivalent to the option not being provided.  Now `null` will indicate that no body is desired.  Cases for consideration:

```
{ "body": undefined }                                               => proxiedBody = req.body; proxiedContentType = req.headers["content-type"]
{ "body": null }                                                           => proxiedBody = "" // e.g. empty body
{ "body": "null", contentType: "application/json" } => proxiedBody = "null"; proxiedContentType = "application/json" // if you want to actually send JSON null
{ "body": "" }                                                              => proxiedBody = '""'; proxiedContentType = "application/json" // 2 byte body
{ "body": "", contentType: "text/plain" }                   => proxiedBody = ""; proxiedContentType = "text/plain" // empty body
```

mcollina (approved):
lgtm

PR #266, opened 2022-08-11
fix: wrong FastifyReplyFromOptions.http type, could be boolean

ACHP:
The documentation specifies here https://github.com/fastify/fastify-reply-from#http that the http props can be a boolean

This is confirmed by this line of code
https://github.com/fastify/fastify-reply-from/blob/5b16f63f6de321f6a5be16b9e89b74182d47092f/lib/request.js#L256

But not in the type definition


RafaelGSS (approved):
LGTM.

PR #268, opened 2022-08-11
fix readme regarding http option, make getHttpOpts resilient against null parameter

Uzlopak:
follow up to #267  and #266 

#### Checklist

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -253,7 +253,7 @@ function getHttp2Opts (opts) {
 }
 
 function getHttpOpts (opts) {
-  const httpOpts = typeof opts.http === 'object' ? opts.http : {}
+  const httpOpts = typeof opts.http === 'object' && opts.http !== null ? opts.http : {}
   httpOpts.requestOptions = httpOpts.requestOptions || {}
 
   if (!httpOpts.requestOptions.timeout) {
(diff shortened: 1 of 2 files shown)


lib/request.js
```diff
@@ -253,7 +253,7 @@ function getHttp2Opts (opts) {
 }
 
 function getHttpOpts (opts) {
-  const httpOpts = typeof opts.http === 'object' ? opts.http : {}
+  const httpOpts = typeof opts.http === 'object' && opts.http !== null ? opts.http : {}
```
mcollina, line 256: can you add a test for this?

Uzlopak: yeah definetly. seems like there are some more issues.

RafaelGSS:
Honestly, I'd follow the `http2` use case https://github.com/fastify/fastify-reply-from/pull/267#issuecomment-1212102941.

IMO It's way clear

mcollina (approved):
lgtm

lib/request.js
Uzlopak: I removed that check, as it seems that null is not the same result as when having an empty object. I assume that it has something to do with how agentOptions are affected.

Uzlopak:
@RafaelGSS 

I think that would result in a breaking change.

PR #272, opened 2022-09-09
perf: replace querystring with fast-querystring

anonrig:
#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,7 +2,7 @@
 
 const fp = require('fastify-plugin')
 const lru = require('tiny-lru')
-const querystring = require('querystring')
+const querystring = require('fast-querystring')
 const Stream = require('stream')
 const buildRequest = require('./lib/request')
 const {
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
   "dependencies": {
     "@fastify/error": "^3.0.0",
     "end-of-stream": "^1.4.4",
+    "fast-querystring": "^1.0.0",
     "fastify-plugin": "^4.0.0",
     "pump": "^3.0.0",
     "tiny-lru": "^8.0.2",


mcollina (approved):
lgtm

Fdawgs (changes requested):
I think we should wait until this is at a stable version (v1.x) before doing this. From the rapid progress @anonrig is making I don't think that's far off!

mcollina (changes requested):
Could you please add an option to configure the querystring module to use? Like in Fastify.

anonrig:
Fastify was using `querystringParser` and this package is using stringifier. I thought of `querystringStringifier` but I'm open to suggestions @Fdawgs & @mcollina

mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test?

anonrig:
I removed the last commit, this is actually implemented already @mcollina https://github.com/fastify/fastify-reply-from#querystring-or-querystringsearch-requrl

mcollina (approved):
lgtm did you check how much this improved perf (if at all)?

anonrig:
@mcollina I didn't do a benchmark, but `fast-querystring`.stringify is 20% faster than `querystring`

Uzlopak:
Can we decorate fastify with querystring and the plugins consume that instead of clustering the option everywhere?

anonrig:
@Fdawgs  I released 1.0. I think it is stable enough.

Fdawgs (approved):

Eomm:
Do we agree to release it as a minor version?

anonrig:
It is a non-breaking change, if we are following sem-ver.

PR #275, opened 2022-10-06
improve test coverage

Uzlopak:
#### Checklist

### DIFF
--- a/.taprc
+++ b/.taprc
@@ -2,10 +2,10 @@ ts: false
 jsx: false
 flow: false
 
-functions: 100
-lines: 99
-statements: 99
-branches: 97
+functions: 97
+lines: 96
+statements: 96
+branches: 96
 
 files:
   - test/**/*.js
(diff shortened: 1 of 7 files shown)


Eomm (commented):
Could you update the .taprc as well?

Uzlopak:
Done. For 100% coverage we would need to test this line
https://github.com/fastify/fastify-reply-from/blob/d4949f4a9697d84186191000f4a5859d5ae3fecd/lib/request.js#L225

I have no clue how this can be tested.

Uzlopak:
Strange. Seems like in older node the coverage is not given? Is it because node14 has no http2 support or so?

Eomm:
There are some bypass:

https://github.com/fastify/fastify-reply-from/blob/eb049b1c22b8372225d22e5056020ece7db43886/test/unix-http-undici-from.js#L11

I would be ok to lower the values for now, we can open an issue on fastify-workflow to manage these use cases

Uzlopak:
Lowered the threshold

PR #277, opened 2022-10-17
fix: verify reply.from() return type

theneva:
Hi! :wave:

I noticed that `reply.from(…)`'s return type is `void`, while in reality it returns the `reply` (like most/all other plugins that decorate the `reply`).

I updated the type, and added a type check to one of the existing tests to verify the fix.

Do you think this looks reasonable?

#### Checklist

- [ ] documentation is changed or added **(not applicable, I think)**
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md) **(I think – I'm not 100 % sure what exactly this entails)**

### DIFF
--- a/types/index.test-d.ts
+++ b/types/index.test-d.ts
@@ -60,7 +60,8 @@ async function main() {
   server.register(replyFrom, fullOptions);
 
   server.get("/v1", (request, reply) => {
-      reply.from();
+      // Verify the return type of reply.from()
+      const newReply: FastifyReply = reply.from();
   });
   server.get("/v3", (request, reply) => {
       reply.from("/v3", {
--- a/types/index.test-d.ts
+++ b/types/index.test-d.ts
@@ -60,8 +60,7 @@ async function main() {
   server.register(replyFrom, fullOptions);
 
   server.get("/v1", (request, reply) => {
-      // Verify the return type of reply.from()
-      const newReply: FastifyReply = reply.from();
+      expectType<FastifyReply>(reply.from());
   });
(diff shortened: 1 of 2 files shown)


types/index.test-d.ts
```diff
@@ -60,7 +60,8 @@ async function main() {
   server.register(replyFrom, fullOptions);
 
   server.get("/v1", (request, reply) => {
-      reply.from();
+      // Verify the return type of reply.from()
+      const newReply: FastifyReply = reply.from();
```
theneva, line 64: I think the code comment is helpful here, since this is the only place that checks the return type of `reply.from(…)`.

Since the code just adds a type verification, and is functionally identical, I figure it's fine to just latch on to an existing handler instead of adding a new one. Let me know if you disagree, though!

theneva:
Oh, by the way, I guess the pull request template comes from https://github.com/fastify/fastify, but there's no benchmark script in this repo, so `npm run benchmark` doesn't work even though the checklist tells you to run it.

Not sure if there' s a reasonable way to alter the template, but perhaps adding a script like `"benchmark": "echo 'This project has no benchmarks!'",` would make contributing slightly less confusing?

types/index.test-d.ts
mcollina: please use an assertion from tsd here?

theneva: Ah, that's way better, thanks! I think d9df225 fixed it :smile:

SimenB (approved):

mcollina (approved):
lgtm

PR #289, opened 2023-02-02
fix: undici headers

### LINKED ISSUE #287: Header encoding changes, leading to `ERR_INVALID_CHAR`
### Prerequisites

- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.12.0

### Plugin version

8.3.1

### Node.js version

16.19.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

13.1

### Description

When a header is set using a character like `å` (Norwegian character) its encoding gets messed up. A fastify server serving it serializes correctly, but if `reply-from` is used, the encoding changes. If another server with `reply-from` is chained, the app crashes with `TypeError [ERR_INVALID_CHAR]: Invalid character in header content ["content-disposition"]`.

### Steps to Reproduce

```js
const replyFrom = require('@fastify/reply-from');
const fastify = require('fastify');

const app = fastify();

app.get('/', (_request, reply) => {
  reply.header('content-disposition', 'år.pdf').send('OK');
});

app.listen({ port: 6666 });

const proxy = fastify();

proxy.register(replyFrom);

proxy.get('/', (_request, reply) => {
  return reply.from('http://localhost:6666');
});

proxy.listen({ port: 7777 });

const secondProxy = fastify();

secondProxy.register(replyFrom);

secondProxy.get('/', (_request, reply) => {
  return reply.from('http://localhost:7777');
});

secondProxy.listen({ port: 8888 });
```

```sh-session
$ curl -i http://localhost:6666
HTTP/1.1 200 OK
content-disposition: år.pdf
content-type: text/plain; charset=utf-8
content-length: 2
Date: Fri, 27 Jan 2023 15:47:49 GMT
Connection: keep-alive
Keep-Alive: timeout=72


$ curl -i http://localhost:7777
HTTP/1.1 200 OK
content-disposition: r.pdf
content-type: text/plain; charset=utf-8
content-length: 2
date: Fri, 27 Jan 2023 15:47:52 GMT
connection: keep-alive
keep-alive: timeout=72


$ curl -i http://localhost:8888
curl: (52) Empty reply from server
```

### Expected Behavior

The header should be correctly encoded and decoded (i.e. `7777` and `8888` should have the same result as `6666`).

If that's not possible I expect the first server to complain when the header is added.

climba03003:
Fixes #287 

Finally found the problem is caused by both `http` and `undici`.
1. `http` is not always treats the `header` to `latin1`.
https://github.com/nodejs/node/blob/2a29df64645a70bbb833298423a29206c4ec6a2e/lib/_http_outgoing.js#L361-L373
2. `undici` always perform `Buffer.from('').toString('utf8')` regardless the `header` encoding.
https://github.com/nodejs/undici/blob/2b260c997ad4efe4ed2064b264b4b546a59e7a67/lib/core/util.js#L216-L229

In combine of the above two problem.
The headers will parsed as a completely difference byte when chaining.

In this fix, we always treats the header as `latin1`. 
So, we are always consistence in the encoding.

#### Checklist

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -12,6 +12,31 @@ function filterPseudoHeaders (headers) {
   return dest
 }
 
+// http required the header to be encoded with latin1
+// undici will convert the latin1 buffer back to utf8
+// https://github.com/nodejs/undici/blob/2b260c997ad4efe4ed2064b264b4b546a59e7a67/lib/core/util.js#L216-L229
+// after chaining, the header will converted in wrong byte
+// Buffer.from('', 'latin1').toString('utf8') applied
+//
+// in order to presist the encoding, always encode it
+// back to latin1
+function patchUndiciHeaders (headers) {
+  const headersKeys = Object.keys(headers)
+  const dist = {}
+
+  let header
+  let i
+
+  for (i = 0; i < headersKeys.length; i++) {
+    header = headersKeys[i]
+    if (header.charCodeAt(0) !== 58) { // fast path for indexOf(':') === 0
+      dist[header] = Buffer.from(headers[header]).toString('latin1')
+    }
+  }
+
+  return dist
+}
+
 function copyHeaders (headers, reply) {
   const headersKeys = Object.keys(headers)
 
@@ -74,6 +99,7 @@ function buildURL (source, reqBase) {
 }
 
 module.exports = {
+  patchUndiciHeaders,
   copyHeaders,
   stripHttp1ConnectionHeaders,
   filterPseudoHeaders,
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -12,13 +12,13 @@ function filterPseudoHeaders (headers) {
   return dest
 }
 
-// http required the header to be encoded with latin1
-// undici will convert the latin1 buffer back to utf8
+// http requires the header to be encoded with latin1
+// undici will convert the latin1 buffer to utf8
 // https://github.com/nodejs/undici/blob/2b260c997ad4efe4ed2064b264b4b546a59e7a67/lib/core/util.js#L216-L229
-// after chaining, the header will converted in wrong byte
+// after chaining, the header will be serialised using wrong encoding
 // Buffer.from('', 'latin1').toString('utf8') applied
 //
-// in order to presist the encoding, always encode it
+// in order to persist the encoding, always encode it
 // back to latin1
 function patchUndiciHeaders (headers) {
   const headersKeys = Object.keys(headers)
--- a/lib/request.js
+++ b/lib/request.js
@@ -5,7 +5,7 @@ const querystring = require('querystring')
 const eos = require('end-of-stream')
 const pump = require('pump')
 const undici = require('undici')
-const { stripHttp1ConnectionHeaders } = require('./utils')
+const { patchUndiciHeaders, stripHttp1ConnectionHeaders } = require('./utils')
 const http2 = require('http2')
 
 const {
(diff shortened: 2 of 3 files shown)


lib/utils.js
```diff
@@ -12,6 +12,31 @@ function filterPseudoHeaders (headers) {
...
 
+// http required the header to be encoded with latin1
+// undici will convert the latin1 buffer back to utf8
+// https://github.com/nodejs/undici/blob/2b260c997ad4efe4ed2064b264b4b546a59e7a67/lib/core/util.js#L216-L229
+// after chaining, the header will converted in wrong byte
+// Buffer.from('', 'latin1').toString('utf8') applied
+//
+// in order to presist the encoding, always encode it
+// back to latin1
+function patchUndiciHeaders (headers) {
```
SimenB, line 23: should there be an option in `undici` to leave headers as `latin1` encoding?

climba03003: I think yes. Otherwise, the `patch` is prone to error.

`undici` always trans-code the header. Maybe it should really detect the encoding before trans-code. That means the behavior should be identical to `http.request`.

Actually, I already open a issue on `undici`.
See https://github.com/nodejs/undici/issues/1903

mcollina (approved):
lgtm

lib/utils.js
```diff
@@ -12,6 +12,31 @@ function filterPseudoHeaders (headers) {
   return dest
 }
 
+// http required the header to be encoded with latin1
```
SimenB, line 15:
```suggestion
// http requires the header to be encoded with latin1
```

lib/utils.js
```diff
@@ -12,6 +12,31 @@ function filterPseudoHeaders (headers) {
...
+// undici will convert the latin1 buffer back to utf8
+// https://github.com/nodejs/undici/blob/2b260c997ad4efe4ed2064b264b4b546a59e7a67/lib/core/util.js#L216-L229
+// after chaining, the header will converted in wrong byte
+// Buffer.from('', 'latin1').toString('utf8') applied
+//
+// in order to presist the encoding, always encode it
```
SimenB, line 21:
```suggestion
// in order to persist the encoding, always encode it
```

lib/utils.js
```diff
@@ -12,6 +12,31 @@ function filterPseudoHeaders (headers) {
   return dest
 }
 
+// http required the header to be encoded with latin1
+// undici will convert the latin1 buffer back to utf8
+// https://github.com/nodejs/undici/blob/2b260c997ad4efe4ed2064b264b4b546a59e7a67/lib/core/util.js#L216-L229
+// after chaining, the header will converted in wrong byte
```
SimenB, line 18:
```suggestion
// after chaining, the header will be serialised using wrong encoding
```

lib/utils.js
```diff
@@ -12,6 +12,31 @@ function filterPseudoHeaders (headers) {
   return dest
 }
 
+// http required the header to be encoded with latin1
+// undici will convert the latin1 buffer back to utf8
```
SimenB, line 16:
```suggestion
// undici will convert the latin1 buffer to utf8
```

RafaelGSS (approved):

mcollina:
Would this need a revert after undici is fixed?

climba03003:
I believe, yes.

arturkin:
FYI this causes regression with `set-cookie` headers in case it's an array of values. Had to revert to the previous version due to this.

climba03003:
> FYI this causes regression with set-cookie headers in case it's an array of values. Had to revert to the previous version due to this.

Can you provide a reproducible code in a separate issue?

SimenB:
The change can probably be reverted together with a bump of the `undici` dependency?

climba03003:
> The change can probably be reverted together with a bump of the undici dependency?

Even with bumping `undici` in `dependencies`, I believe it wouldn't catch-up the latest.

SimenB:
It would

mcollina:
Could you open a fresh PR?

Issue #168, opened 2021-04-23
Using `rewriteRequestHeaders` with async/await does not work properly


vorillaz:
## 🐛 Bug Report

It seems that when using an `async` callback for `rewriteRequestHeaders` an empty object gets returned and the API overrides the headers. Original I found bug while using https://github.com/fastify/fastify-http-proxy

## To Reproduce

```js
const geoip = require("fast-geoip");
const Fastify = require("fastify");

const target = Fastify({
	logger: true,
});

target.get("/", (request, reply) => {
  console.log(request.headers);
	reply.send("hello world");
});

const proxy = Fastify({
	logger: true,
});

proxy.register(require("fastify-reply-from"), {
	base: "http://localhost:3001/",
	rewriteRequestHeaders: async (request, headers) => {
		const {country} = await geoip.lookup("192.11.111.111");
                return {...headers, country, time: Date.now()}
	},
});
```

## Expected behavior
Expecting to receive the headers from the upstream along with `country` and `time` headers

## Your Environment

- *node version*: 14
- *fastify version*: latest
- *os*: Mac

mcollina:
This function is (currently) synchronous. I would recommend you to use a preHandler hook to implement geolocation instead and decorate the request.

PR #293, opened 2023-02-22
Pass Fastify request to callbacks

### LINKED ISSUE #72: `rewriteRequestHeaders` unable to access request decorations
## 🐛 Bug Report

`rewriteRequestHeaders` unable to access request decorations

## To Reproduce

```js
fastify.decorateRequest('foo', null)
fastify.addHook('preHandler', async (request, reply) => {
  request.foo = 'Bar'
})
fastify.register(require('fastify-http-proxy'), {
  upstream: process.env.UPSTREAM,
  prefix: '/api',
  replyOptions: {
    rewriteRequestHeaders(originalReq, headers) {
      const modifiedHeaders = {
        ...headers,
        'x-foo': request.foo
      }
      return modifiedHeaders
    }
  }
})
```

## Expected behavior

I expect that the request decorations would be accessible via `rewriteRequestHeaders` via the `originalReq` parameter.

## Your Environment

- *node version*: 12
- *fastify version*: 2.13.1
- *os*: Mac

barronpm:
Fixes #72, related to #168.

Currently, the options `rewriteHeaders`, `rewriteRequestHeaders`, and `getUpstream` are all sync-only.
The suggested workaround in #168 was to use hooks to decorate the request, however, only `onResponse` passes in the fastify request, the others passed in the raw request, rendering the decorations inaccessible.

This PR updates the methods above to use the fastify request as the parameter. The old parameter is still accessible at `request.raw`.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,20 +296,20 @@ It must reply the error.
 
 Called to rewrite the headers of the response, before them being copied
 over to the outer response.
-Parameters are the original headers and the fastify request.
+Parameters are the original headers and the Fastify request.
 It must return the new headers object.
 
 #### `rewriteRequestHeaders(request, headers)`
 
 Called to rewrite the headers of the request, before them being sent to the other server.
-Parameters are the fastify request and the original request headers.
+Parameters are the Fastify request and the original request headers.
 It must return the new headers object.
 
 #### `getUpstream(request, base)`
 
 Called to get upstream destination, before the request is being sent. Useful when you want to decide which target server to call based on the request data.
 Helpful for a gradual rollout of new services.
-Parameters are the fastify request and the base string from the plugin options.
+Parameters are the Fastify request and the base string from the plugin options.
 It must return the upstream destination.
 
 #### `queryString` or `queryString(search, reqUrl)`
--- a/README.md
+++ b/README.md
@@ -292,21 +292,24 @@ The default behavior is `reply.send(error)`, which will be disabled if the
 option is specified.
 It must reply the error.
 
-#### `rewriteHeaders(headers, req)`
+#### `rewriteHeaders(headers, request)`
 
 Called to rewrite the headers of the response, before them being copied
 over to the outer response.
+Parameters are the original headers and the fastify request.
 It must return the new headers object.
 
-#### `rewriteRequestHeaders(originalReq, headers)`
+#### `rewriteRequestHeaders(request, headers)`
 
 Called to rewrite the headers of the request, before them being sent to the other server.
+Parameters are the fastify request and the original request headers.
 It must return the new headers object.
 
-#### `getUpstream(originalReq, base)`
+#### `getUpstream(request, base)`
 
 Called to get upstream destination, before the request is being sent. Useful when you want to decide which target server to call based on the request data.
 Helpful for a gradual rollout of new services.
+Parameters are the fastify request and the base string from the plugin options.
 It must return the upstream destination.
 
 #### `queryString` or `queryString(search, reqUrl)`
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -63,7 +63,7 @@ const fastifyReplyFrom = fp(function from (fastify, opts, next) {
     }
 
     // we leverage caching to avoid parsing the destination URL
-    const dest = getUpstream(req, base)
+    const dest = getUpstream(this.request, base)
```
mcollina, line 66: We need a test for this change.

barronpm: There's already a test checking the type, is there something else you wanted to test for?

jsumners: If the current tests do not show a problem then they do not cover the change.

barronpm: I had to edit the tests to check for the new type, see my edits to `types/index.test-d.ts`

jsumners: This is a change to the actual code. There needs to be an associated unit test to guard against regression and assure that the data is being passed correctly.

barronpm: Alright, how do you want me to check whether the argument is a fastify request?

jsumners: That's the core of the work. How did you determine that it needs to be passed in? Write a test that fails your expectation without the change.

Eomm: right now: `req` is equals to `this.request.raw`, `this.request` is the fastify request.

In any case:

- this PR would be a breaking change
- we need to update docs
- tests must be updated to check that option functions get a fastify object

barronpm: I've added tests that check that the request passed in the callback is correct, and verified that those tests fail on master.

In terms of docs, I read through the README, but the docs never specified exactly what the type of the request parameter is, so I left it as-is. Do you want me to add that?

mcollina: Please add it, thanks!

barronpm: done

README.md
```diff
@@ -292,21 +292,24 @@ The default behavior is `reply.send(error)`, which will be disabled if the
 option is specified.
 It must reply the error.
 
-#### `rewriteHeaders(headers, req)`
+#### `rewriteHeaders(headers, request)`
 
 Called to rewrite the headers of the response, before them being copied
 over to the outer response.
+Parameters are the original headers and the fastify request.
```
jsumners, line 299:
```suggestion
Parameters are the original headers and the Fastify request.
```

README.md
```diff
@@ -292,21 +292,24 @@ The default behavior is `reply.send(error)`, which will be disabled if the
...
 It must return the new headers object.
 
-#### `rewriteRequestHeaders(originalReq, headers)`
+#### `rewriteRequestHeaders(request, headers)`
 
 Called to rewrite the headers of the request, before them being sent to the other server.
+Parameters are the fastify request and the original request headers.
```
jsumners, line 305:
```suggestion
Parameters are the Fastify request and the original request headers.
```

README.md
```diff
@@ -292,21 +292,24 @@ The default behavior is `reply.send(error)`, which will be disabled if the
...
 It must return the new headers object.
 
-#### `getUpstream(originalReq, base)`
+#### `getUpstream(request, base)`
 
 Called to get upstream destination, before the request is being sent. Useful when you want to decide which target server to call based on the request data.
 Helpful for a gradual rollout of new services.
+Parameters are the fastify request and the base string from the plugin options.
```
jsumners, line 312:
```suggestion
Parameters are the Fastify request and the base string from the plugin options.
```

mcollina (approved):
lgtm

PR #299, opened 2023-03-09
Disable http/http2 timeout for sse events

### LINKED ISSUE #298: Option to disable request timeout
### Prerequisites

- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Option to completely disable the request timeout (i.e. not pass anything in [`requestOptions.timeout`](https://github.com/fastify/fastify-reply-from/blob/master/lib/request.js#L268)).

### Motivation

We have an app that uses SSE, but it doesn't send requests directly to that server. It sends them to another server that uses `fastify-reply-from` to forward the request to the server that sends the events. Due to the persistent http connection, we'll always get a timeout error after whatever was configured for the `opts.http.requestOptions.timeout`.

We want to disable the timeout in this package because we still want to use this proxy requests to a from one server to another.

### Example

To avoid a breaking change, perhaps `timeout` could also be a `boolean`. So code that looks like below. Other options like allowing `null` to mean no timeout are there, but may break someone's code.

```js
// request.js line 267
if (
  // if timeout is not a boolean and falsy, use the default. 
  // if timeout is a boolean, use the default only if it's true
  typeof httpOpts.requestOptions.timeout !== 'boolean' && !httpOpts.requestOptions.timeout ||
  httpOpts.requestOptions.timeout
) {
  httpOpts.requestOptions.timeout = 10000
}
```
```js
// example usage
proxy.register(require('@fastify/reply-from'), {
  base: 'http://localhost:3001/'
  http: {
    requestOptions: {
      timeout: false
    }
  }
})
```

deanhaleem:
Related to #298. It was mentioned there `undici` possibly has a bug regarding this. I will work on creating a small reproduction and create an issue on that module

This checks the response headers for `content-type: text/event-stream`, and if found, get rid of any existing timeouts.

Fixes #298

#### Checklist

### DIFF
--- a/test/http-timeout.test.js
+++ b/test/http-timeout.test.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const t = require('tap')
+const { test } = require('tap')
 const Fastify = require('fastify')
 const From = require('..')
 const got = require('got')
@@ -8,21 +8,21 @@ const FakeTimers = require('@sinonjs/fake-timers')
 
 const clock = FakeTimers.createClock()
 
-t.autoend(false)
+test('http request timeout', async (t) => {
+  t.autoend(false)
 
-const target = Fastify()
-t.teardown(target.close.bind(target))
+  const target = Fastify()
+  t.teardown(target.close.bind(target))
 
-target.get('/', (request, reply) => {
-  t.pass('request arrives')
+  target.get('/', (request, reply) => {
+    t.pass('request arrives')
 
-  clock.setTimeout(() => {
-    reply.status(200).send('hello world')
-    t.end()
-  }, 200)
-})
+    clock.setTimeout(() => {
+      reply.status(200).send('hello world')
+      t.end()
+    }, 200)
+  })
 
-async function main () {
   await target.listen({ port: 0 })
 
   const instance = Fastify()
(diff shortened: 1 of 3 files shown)


test/http-timeout.test.js
```diff
@@ -52,6 +52,31 @@ async function main () {
...
-main()
+test('http sse removes timeout test', async (t) => {
+  const target = Fastify()
+  t.teardown(target.close.bind(target))
+
+  target.get('/', (request, reply) => {
+    t.pass('request arrives')
+
+    reply.header('content-type', 'text/event-stream').status(200).send('hello world')
+  })
```
deanhaleem, line 65: I admit I was struggling a little figuring out the best way to test these changes. The way this works is it more tests the code via the coverage: without the new tests, the coverage lowers.

If it's preferred to instead specifically test that `req.setTimeout(0)` is being called, for sure can let me know

mcollina (approved):
lgtm

deanhaleem:
> Can you please open an issue on undici showing that the request is timing out if there is a SSE response? So we can track it upstream.
> 
> I think it would be better to have the http2 changes here.

Thanks, added the http2 changes back 👍  realized the initial reason it didn't work was missing the `http2Client` timeout.

And yes I will add an issue, need to find time to create a small reproduction first to make sure it's not an issue in my application.

mcollina (approved):
lgtm

PR #300, opened 2023-03-12
Fix default http2 timeout to be 60 seconds

deanhaleem:
Noticed in the readme it's stated the default timeout for http2 requests is 60 seconds, but in the code it's 6 seconds. If it was actually the readme that was incorrect, will be more than happy to fix it there instead

#### Checklist

### DIFF
--- a/test/http2-target-multi-crash.test.js
+++ b/test/http2-target-multi-crash.test.js
@@ -16,7 +16,9 @@ test('http -> http2 crash multiple times', async (t) => {
 
   instance.register(From, {
     base: 'http://localhost:3128',
-    http2: true,
+    http2: {
+      sessionTimeout: 6000
+    },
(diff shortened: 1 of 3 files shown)


test/http2-target-multi-crash.test.js
```diff
@@ -16,7 +16,9 @@ test('http -> http2 crash multiple times', async (t) => {
 
   instance.register(From, {
     base: 'http://localhost:3128',
-    http2: true,
+    http2: {
+      sessionTimeout: 6000
+    },
```
deanhaleem, line 21: The tests updates are so these tests work the same as they did before (i.e. use the 6 second timeout rather than the 60 second one)

mcollina (approved):
lgtm

PR #307, opened 2023-04-08
updated readme that getUpStream is not working with http2 and added a…

nickwinger:
…n example

just updated the readme that getUpStream is not working with http2.
Note:
I could (correctly) also update the code in `decorateReply`, like this:
```
if (!!opts.getUpstream && !!opts.http2)
      throw new Error('getUpStream with http2 not yet supported. See readme for using constraintStrategy instead')
```

but that would make a breaking-change, so i guess one would have to do that at another release-version

#### Checklist


mcollina (approved):
lgtm

PR #310, opened 2023-05-05
feat: pass full response object to onResponse hook

airhorns:
Before this change, the onResponse hook was only given the response stream to work with. This was useful if you wanted to transform the response body on a chunk by chunk basis, but did not allow inspecting the status code or headers of the response! If you wanted to serve a custom error page or similar, you couldn't get at the actual `res` object in question.

So, this changes the `onResponse` hook to instead pass the whole response instead of just the stream. Now, users can inspect the status code as well.

This is a breaking change and would thus be semver major. I feel like its an advanced enough use case that the churn won't be super high, but if we hate the idea of a breaking change, we could instead add a forth parameter to the `onResponse` hook which is the `res` object. Bit of a weird API so I went for the cleanest one.

#### Checklist


mcollina:
can you target the `next` branch? this is a breaking change and we'll need to land it in v5.

airhorns:
There is no next branch in this repo -- shall I create one branched from main?

Uzlopak:
I thought you were not a maintainer and created a next branch from master. Well, now we have a next branch.

mcollina:
Seems that CI is failing.

Uzlopak:
The CI is generally failing. It is not related to this PR.

mcollina:
I'm a little bit worried of shipping a breaking change with a failing CI to begin with.

Fdawgs:
CI failure is related to node 20 being added to the shared CI workflow, which this repo uses, and something isn't playing ball with node 20.

Haven't had time to investigate due to starting a new job last week.

Fdawgs:
> CI failure is related to node 20 being added to the shared CI workflow, which this repo uses, and something isn't playing ball with node 20.
> 
> Haven't had time to investigate due to starting a new job last week.

Test failure appears to be related to Node 19 changing the default value for the `connection` header from `close` to `keep-alive`, see https://nodejs.org/en/blog/release/v19.0.0#https11-keepalive-by-default.

Not too sure what change should be made to the tests to accommodate this, skip testing the header for node >= 19 like  https://github.com/fastify/fastify/blob/c4cb3a87cdbd33524556314341aa2b9bea61422f/test/close.test.js#L208 ?

Uzlopak:
@Fdawgs 
provided #312

mcollina:
@Uzlopak could you rebase? thx

airhorns:
I'll do it, I think I need to update some types as well!

airhorns:
Of note: the types for this argument were broken on main and already said it was the full raw response object being passed, not the stream -- so this PR actually fixes the types! Added type tests to cover this.

mcollina (approved):
lgtm

PR #322, opened 2023-06-15
feat: add `method` opt to overwrite http method

psteinroe:
adds a `method` opt to overwrite the http method of the request that is being sent to the target. 

use case is that the target may only support a `POST` request, for example because it does not confirm to certain standards, but the proxy should expose the same endpoint as `PATCH`. with the `method` parameter set to `POST`, the incoming `PATCH` request is forwarded as `POST`.

#### Checklist

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -67,6 +67,7 @@ declare namespace fastifyReplyFrom {
       request: FastifyRequest<RequestGenericInterface, RawServerBase>,
       base: string
     ) => string;
+    method?: HTTPMethods;
   }
 
   interface Http2Options {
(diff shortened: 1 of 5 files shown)


types/index.d.ts
```diff
@@ -67,6 +67,7 @@ declare namespace fastifyReplyFrom {
       request: FastifyRequest<RequestGenericInterface, RawServerBase>,
       base: string
     ) => string;
+    method?: HTTPMethods;
```
mcollina, line 70: Can you add a test for this new prop in the types? We use `tsd`.

mcollina (approved):
lgtm

psteinroe:
> lgtm

Thanks! Anything left to do so this can be merged and released?

PR #332, opened 2023-11-06
[feature] custom retry callback

MikePresman:
# This PR implements two fundamental changes.
Edit: ~~1. I have added `PATCH` and `POST` as given default `retryMethods`~~ see review by airhorns
2. I have added the option to pass a callback to handle retries in a custom fashion as defined by the client.

## Point 2. Described
The motivation for this change is that out of the box `reply-from` does not retry 500's so gracefully handling such cases is not possible. As a result with this change you are able to write your own callback to retry 500's as you wish (or any other kind of error - using 500's here as an exemplar).

This can be done quite simply.
 ```
 const customRetryLogic = (req, res, registerDefaultRetry, defaultRetryAfter) => {
    if (res && res.statusCode === 500 && req.method === 'GET') {
      return 300
    }
    return null
  }

...
Fastify.reply(From, customRetry: { handler: customRetryLogic, retries: 10 }}
```

Note this new API is an object which has two keys. `handler` and `retries` where handler is a required key and retries is optional but strongly encouraged.

-----
### There is an important intricacy to this API.
If no `customRetry` object is passed then business as usual in the library. Retries work as they are expected for default cases.

But if a `customRetry` is passed then it is up to responsibility of the callback to be aware that if they expect default behaviour to co-exist they must register it. If they don't (the above example DID NOT) all default cases of errors (such as 503s) won't be retried.

Example to get default behaviour in a `customRetry`

 ```
 const customRetryLogic = (req, res, registerDefaultRetry, defaultRetryAfter) => {
    // Registering the default retry logic so if a default case occurs we handle it.
    if (registerDefaultRetry()){
      return defaultRetryAfter; // Some new magic is that we can override our headers and pass a new retry-after value.
    }

    if (res && res.statusCode === 500 && req.method === 'GET') {
      return 300
    }
    return null
  }
```

CC: @airhorns 

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,8 @@ npm i @fastify/reply-from
 ```
 
 ## Compatibility with @fastify/multipart
-`@fastify/reply-from` and [`@fastify/multipart`](https://github.com/fastify/fastify-multipart) should not be registered as sibling plugins nor should they be registered in plugins which have a parent-child relationship.<br> The two plugins are incompatible, in the sense that the behavior of `@fastify/reply-from` might not be the expected one when the above-mentioned conditions are not respected.<br> This is due to the fact that `@fastify/multipart` consumes the multipart content by parsing it, hence this content is not forwarded to the target service by `@fastify/reply-from`.<br>
+
+`@fastify/reply-from` and [`@fastify/multipart`](https://github.com/fastify/fastify-multipart) should not be registered as sibling plugins nor should they be registered in plugins which have a parent-child relationship.`<br>` The two plugins are incompatible, in the sense that the behavior of `@fastify/reply-from` might not be the expected one when the above-mentioned conditions are not respected.`<br>` This is due to the fact that `@fastify/multipart` consumes the multipart content by parsing it, hence this content is not forwarded to the target service by `@fastify/reply-from`.`<br>`
 However, the two plugins may be used within the same fastify instance, at the condition that they belong to disjoint branches of the fastify plugins hierarchy tree.
 
 ## Usage
@@ -102,6 +103,7 @@ proxy.register(require('@fastify/reply-from'), {
   }
 })
 ```
+
 See undici own options for more configurations.
 
 You can also pass the plugin a custom instance:
@@ -136,6 +138,7 @@ proxy.register(require('@fastify/reply-from'), {
 ```
 
 You can also pass custom HTTP agents. If you pass the agents, then the http.agentOptions will be ignored. To illustrate:
+
 ```js
 proxy.register(require('@fastify/reply-from'), {
   base: 'http://localhost:3001/',
@@ -221,6 +224,7 @@ This plugin will always retry on 503 errors, _unless_ `retryMethods` does not co
 Enables the possibility to explictly opt-in for global agents.
 
 Usage for undici global agent:
+
 ```js
 import { setGlobalDispatcher, ProxyAgent } from 'undici'
 
@@ -234,6 +238,7 @@ fastify.register(FastifyReplyFrom, {
 ```
 
 Usage for http/https global agent:
+
 ```js
 fastify.register(FastifyReplyFrom, {
   base: 'http://localhost:3001/',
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -251,29 +252,51 @@ function isFastifyMultipartRegistered (fastify) {
...
+
+          if (res && res.headers['retry-after']) {
+            retryAfter = res.headers['retry-after']
+          }
+          return retryAfter
         }
-        if (!reply.sent) {
-          // always retry on 503 errors
+
+        const defaultRetry = () => {
```
airhorns, line 272: Why two separate functions for the duration and if we should retry or not? I think it'd be simpler with one function that returns null or a number, null meaning don't retry, and a number meaning retry after that many milliseconds.

README.md
```diff
@@ -264,6 +264,37 @@ This option set the limit on how many times the plugin should retry the request,
...
+    if (registerDefaultRetry()){
+      return defaultRetryAfter;
+    }
+
+    //Custom retry logic
+    if (res && res.statusCode === 500 && req.method === 'GET') {
+      return 300
+    }
+    return null
+  }
```
airhorns, line 288: I think it might be clearer not to describe this as "registering" another retry handler (both in this PR and in the test code), we're not really adding a handler to an event emitter or something, but just passing an option. 

Also, I think that a custom retry handler needs to be passed the number of attempts so far if it wants to calculate an exponential delay. What do you think about this API instaead

```suggestion
const customRetryLogic = (req, res, attemptCount, getDefaultDelay) => {

    const defaultDelay = getDefaultDelay();
    if (defaultDelay) return defaultDelay();
    
    if (res && res.statusCode === 500 && req.method === 'GET') {
      return 300
    }
    return null
  }
```

index.js
```diff
@@ -251,29 +252,51 @@ function isFastifyMultipartRegistered (fastify) {
   return (fastify.hasContentTypeParser('multipart') || fastify.hasContentTypeParser('multipart/form-data')) && fastify.hasRequestDecorator('multipart')
 }
 
-function createRequestRetry (requestImpl, reply, retriesCount, retryOnError, maxRetriesOn503) {
+function createRequestRetry (requestImpl, reply, retriesCount, retryOnError, maxRetriesOn503, customRetry) {
```
airhorns, line 255: The method signature to this function seems a little complicated. I think it might be simpler if the caller just built one "give me the retry delay" function, and passed it into this thing. That function can do whatever it needs to based on the options, and kind of bottles up all the logic, so this thing doesn't really need to care about the "when" of retrying, and is just about actually doing it based on the output of that function.

test/retry-with-a-custom-handler.test.js
```diff
@@ -0,0 +1,114 @@
...
+
+  try {
+    await got.get(`http://localhost:${instance.server.address().port}`, { retry: 0 })
+  } catch (error) {
+    t.equal(error.message, 'Response code 503 (Service Unavailable)')
+    t.end()
+  }
+})
+
+test("a server 503's with a custom handler for 500 and the custom handler registers the default so it passes", async (t) => {
```
airhorns, line 93: If you are down with the language change
```suggestion
test("custom retry delay functions can invoke the default delay", async (t) => {
```

mcollina (approved):
lgtm

airhorns (approved):
LGTM after comments!

README.md
```diff
@@ -14,7 +14,7 @@ npm i @fastify/reply-from
...
-`@fastify/reply-from` and [`@fastify/multipart`](https://github.com/fastify/fastify-multipart) should not be registered as sibling plugins nor should they be registered in plugins which have a parent-child relationship.<br> The two plugins are incompatible, in the sense that the behavior of `@fastify/reply-from` might not be the expected one when the above-mentioned conditions are not respected.<br> This is due to the fact that `@fastify/multipart` consumes the multipart content by parsing it, hence this content is not forwarded to the target service by `@fastify/reply-from`.<br>
+`@fastify/reply-from` and [`@fastify/multipart`](https://github.com/fastify/fastify-multipart) should not be registered as sibling plugins nor should they be registered in plugins which have a parent-child relationship.`<br>` The two plugins are incompatible, in the sense that the behavior of `@fastify/reply-from` might not be the expected one when the above-mentioned conditions are not respected.`<br>` This is due to the fact that `@fastify/multipart` consumes the multipart content by parsing it, hence this content is not forwarded to the target service by `@fastify/reply-from`.`<br>`
```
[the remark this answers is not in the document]
MikePresman: I think I ran lint incorrectly my bad !
Will fix.

PR #337, opened 2023-12-19
add: error object to custom retry callback

MikePresman:
This change introduces passing the error object to the custom retry callback; something that should've been done in the initial implementation of this feature but forgotten. Alas this PR introduces passing the error object to the callback.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,18 +268,25 @@ By Default: 10
 
 ---
 
-### `customRetry`
+### `retryDelay`
 
 - `handler`. Required
-- `retries`. Optional
 
-This plugin gives the client an option to pass their own retry callback to handle retries on their own.
-If a `handler` is passed to the `customRetry` object the onus is on the client to invoke the default retry logic in their callback otherwise default cases such as 503 will not be handled
+This plugin gives the client an option to pass their own retry callback to allow the client to define what retryDelay they would like on any retries
+outside the scope of what is handled by default in fastify-reply-from. To see the default please refer to index.js `getDefaultDelay()`
+If a `handler` is passed to the `retryDelay` object the onus is on the client to invoke the default retry logic in their callback otherwise default cases such as 500 will not be handled
+
+- The `attempt` property in the object callback refers to the current retriesAttempt number.
+You are given the freedom to use this in concert with the retryCount property set to handle retries
+
+- `getDefaultRetry` refers to the default retry handler. If this callback returns not null and you wish to handle those case of errors simply invoke it as done below.
+
+- `err`, `req`, `res` are self-explanatory error, request, and response properties respectively
 
 Given example
 
 ```js
-   const customRetryLogic = ({req, res, err, getDefaultRetry}) => {
+   const customRetryLogic = ({err, req, res, attempt, getDefaultRetry}) => {
     //If this block is not included all non 500 errors will not be retried
     const defaultDelay = getDefaultDelay();
     if (defaultDelay) return defaultDelay();
@@ -300,7 +307,7 @@ Given example
 
 fastify.register(FastifyReplyFrom, {
   base: 'http://localhost:3001/',
-  customRetry: {handler: customRetryLogic, retries: 10}
+  customRetry: {handler: customRetryLogic}
 })
 
 ```
--- a/README.md
+++ b/README.md
@@ -276,12 +276,11 @@ This plugin gives the client an option to pass their own retry callback to allow
 outside the scope of what is handled by default in fastify-reply-from. To see the default please refer to index.js `getDefaultDelay()`
 If a `handler` is passed to the `retryDelay` object the onus is on the client to invoke the default retry logic in their callback otherwise default cases such as 500 will not be handled
 
-- The `attempt` property in the object callback refers to the current retriesAttempt number.
-You are given the freedom to use this in concert with the retryCount property set to handle retries
-
+- `err` is the error thrown by making a request using whichever agent is configured
+- `req` is the raw request details sent to the underlying agent. __Note__: this object is not a Fastify request object, but instead the low-level request for the agent.
+- `res` is the raw response returned by the underlying agent (if available) __Note__: this object is not a Fastify response, but instead the low-level response from the agent. This property may be null if no response was obtained at all, like from a connection reset or timeout.
+- `attempt` in the object callback refers to the current retriesAttempt number. You are given the freedom to use this in concert with the retryCount property set to handle retries
 - `getDefaultRetry` refers to the default retry handler. If this callback returns not null and you wish to handle those case of errors simply invoke it as done below.
-
-- `err`, `req`, `res` are self-explanatory error, request, and response properties respectively
 
 Given example
 
@@ -307,7 +306,7 @@ Given example
 
 fastify.register(FastifyReplyFrom, {
   base: 'http://localhost:3001/',
-  customRetry: {handler: customRetryLogic}
+  retryDelay: customRetryLogic
 })
 
 ```
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ If a `handler` is passed to the `customRetry` object the onus is on the client t
 Given example
 
 ```js
-   const customRetryLogic = (req, res, err, getDefaultRetry) => {
+   const customRetryLogic = ({req, res, err, getDefaultRetry}) => {
     //If this block is not included all non 500 errors will not be retried
     const defaultDelay = getDefaultDelay();
     if (defaultDelay) return defaultDelay();
@@ -305,6 +305,14 @@ fastify.register(FastifyReplyFrom, {
 
 ```
 
+Note the Typescript Equivalent
+```
+const customRetryLogic = ({req, res, err, getDefaultRetry}: customRetryHandler) => {
+  ...
+}
+...
+
+```
 ---
 
 ### `reply.from(source, [opts])`
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ If a `handler` is passed to the `customRetry` object the onus is on the client t
 Given example
 
 ```js
-   const customRetryLogic = (req, res, getDefaultRetry, err) => {
+   const customRetryLogic = (req, res, err, getDefaultRetry) => {
     //If this block is not included all non 500 errors will not be retried
     const defaultDelay = getDefaultDelay();
     if (defaultDelay) return defaultDelay();
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ If a `handler` is passed to the `customRetry` object the onus is on the client t
 Given example
 
 ```js
-  const customRetryLogic = (req, res, getDefaultRetry) => {
+   const customRetryLogic = (req, res, getDefaultRetry, err) => {
     //If this block is not included all non 500 errors will not be retried
     const defaultDelay = getDefaultDelay();
     if (defaultDelay) return defaultDelay();
@@ -288,6 +288,11 @@ Given example
     if (res && res.statusCode === 500 && req.method === 'GET') {
       return 300
     }
+
+    if (err && err.code == "UND_ERR_SOCKET"){
+      return 600
+    }
+
     return null
   }
 
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ fastify.register(FastifyReplyFrom, {
 
 Note the Typescript Equivalent
 ```
-const customRetryLogic = ({req, res, err, getDefaultRetry}: customRetryHandler) => {
+const customRetryLogic = ({req, res, err, getDefaultRetry}: RetryDetails) => {
   ...
 }
 ...
--- a/index.js
+++ b/index.js
@@ -165,7 +165,7 @@ const fastifyReplyFrom = fp(function from (fastify, opts, next) {
         if (customRetry && customRetry.handler) {
           const customRetries = customRetry.retries || 1
           if (++retries < customRetries) {
-            return customRetry.handler(req, res, defaultDelay, err)
+            return customRetry.handler(req, res, err, defaultDelay)
           }
         }
         return defaultDelay()
--- a/index.js
+++ b/index.js
@@ -165,7 +165,7 @@ const fastifyReplyFrom = fp(function from (fastify, opts, next) {
         if (customRetry && customRetry.handler) {
           const customRetries = customRetry.retries || 1
           if (++retries < customRetries) {
-            return customRetry.handler(req, res, defaultDelay)
+            return customRetry.handler(req, res, defaultDelay, err)
           }
         }
         return defaultDelay()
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -279,7 +279,7 @@ If a `handler` is passed to the `customRetry` object the onus is on the client t
 Given example
 
 ```js
-  const customRetryLogic = (req, res, getDefaultRetry) => {
+   const customRetryLogic = (req, res, getDefaultRetry, err) => {
```
airhorns, line 282: Do you like this signature or `(req, res, err, getDefaultRetry)` better? The original PR hasn't been released yet so I think we can make a breaking change here if we want.

airhorns:
@MikePresman want to just include the types PR in this one and adjust the types, and we can merge it together and close the other one?

types/index.test-d.ts
```diff
@@ -91,6 +91,18 @@ async function main() {
   instance.get("/http2", (request, reply) => {
       reply.from("/", {
           method: "POST",
+          customRetry: {
+            handler: (req, res, getDefaultDelay) => {
```
airhorns, line 95: can you poke at the error here

MikePresman: of course, sorry 🤦

airhorns (approved):

index.js
```diff
@@ -165,7 +165,7 @@ const fastifyReplyFrom = fp(function from (fastify, opts, next) {
         if (customRetry && customRetry.handler) {
           const customRetries = customRetry.retries || 1
           if (++retries < customRetries) {
-            return customRetry.handler(req, res, defaultDelay)
+            return customRetry.handler(req, res, err, defaultDelay)
```
Eomm, line 168: Since it will be braking change, I wonder if an object is better so users will have a smooter transition and we could add more info in future if needed (such as the num of retry) without shipping it as breaking

airhorns: I don't think it is a breaking change because the original PR hasn't been released yet -- it has just been sitting on main!

Eomm: Oh nice!

Do you think the object interface could be better?

airhorns: You mean like a destructuring approach where you can only pull out the properties you want?

Eomm: Yes, this one:

```suggestion
            return customRetry.handler({ req, res, err, defaultDelay })
```

airhorns: I figure that is a little bit less performant as it adds another allocation and restructuring, and IMO 4 args is a lot but its not so many that we need to do this. If we suspected that we were going to add more properties over time and wanted a way to do that without making a breaking change, I'd say that's a great idea, but I don't think there's much more information that is going to exist about an error that we would want to pass in in the future. If you feel strongly I am ok with it.

Eomm: The most common use I see:
after 4 retries I want to log the message, otherwise ignore it.

For this reason I think this object will change soonish

airhorns: Can you explain more? Are you saying we should add the attempt count?

test/retry-with-a-custom-handler.test.js
```diff
@@ -6,10 +6,11 @@ const From = require('..')
 const http = require('node:http')
 const got = require('got')
 
-function serverWithCustomError (stopAfter, statusCodeToFailOn) {
+function serverWithCustomError (stopAfter, statusCodeToFailOn, closesSocket) {
```
Uzlopak, line 9: I would rename it to `closeSocket`

index.js
Uzlopak: If we break, what about putting the error to be in the first place?

test/retry-with-a-custom-handler.test.js
MikePresman: I've made the change but curious if you could expand on why.

mcollina:
There are a lot of CI failures

test/retry-with-a-custom-handler.test.js
Uzlopak: Because closeSocket is imperative and tells the function to do something, while closesSocket is passive indicative or sounds like a question.

index.js
```diff
@@ -165,10 +165,10 @@ const fastifyReplyFrom = fp(function from (fastify, opts, next) {
         if (customRetry && customRetry.handler) {
           const customRetries = customRetry.retries || 1
           if (++retries < customRetries) {
-            return customRetry.handler(req, res, defaultDelay)
+            return customRetry.handler({err, req, res, getDefaultDelay})
```
Uzlopak, line 168:
```suggestion
            return customRetry.handler({ err, req, res, getDefaultDelay })
```

types/index.d.ts
```diff
@@ -39,12 +39,22 @@ declare module "fastify" {
 }
 
 type FastifyReplyFrom = FastifyPluginCallback<fastifyReplyFrom.FastifyReplyFromOptions>
-
+export type customRetryHandler = {
+  err: Error;
+  req: FastifyRequest<RequestGenericInterface, RawServerBase>;
+  res: FastifyReply<RawServerBase>;
+  getDefaultDelay: () => number | null;
+}
```
Uzlopak, line 47: This has to be in the namespace of fastifyReplyFrom

MikePresman:
Apologies about the previous failing ci and lint, had pushed and it slipped my mind to check when I grabbed lunch. All should be well now.

gurgunday (approved):
lgtm, **breaking change**

index.js
Eomm: It is not a breaking change, this feature is not yet released

MikePresman:
> Not a major, this feature is not yet released Here is the discussion [#337 (comment)](https://github.com/fastify/fastify-reply-from/pull/337#discussion_r1432989697)
> 
> I'm not going to approve as my comment is not solved, feel free to land it if you disagree 👍

Could you please clarify, I've made the change to interfacing the function args as an object which is what I believe was the core of your comment unless you meant 
```
The most common use I see:
after 4 retries I want to log the message, otherwise ignore it.

For this reason I think this object will change soonish
```
passing the logger is something you want in this PR.

gurgunday:
> I'm not going to approve as my comment is not solved, feel free to land it if you disagree 👍

I personally wouldn't merge without making sure that we're all in agreement

[Hasn't your proposal been implemented?](https://github.com/fastify/fastify-reply-from/pull/337/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R168)

README.md
```diff
@@ -300,7 +307,7 @@ Given example
 
 fastify.register(FastifyReplyFrom, {
   base: 'http://localhost:3001/',
-  customRetry: {handler: customRetryLogic, retries: 10}
+  customRetry: {handler: customRetryLogic}
```
airhorns, line 310: Can we avoid this options object taking only one property, and just accept the function directly? Sorry to be annoying!
```suggestion
  retryDelay: customRetryLogic
```

README.md
```diff
@@ -268,18 +268,25 @@ By Default: 10
...
+This plugin gives the client an option to pass their own retry callback to allow the client to define what retryDelay they would like on any retries
+outside the scope of what is handled by default in fastify-reply-from. To see the default please refer to index.js `getDefaultDelay()`
+If a `handler` is passed to the `retryDelay` object the onus is on the client to invoke the default retry logic in their callback otherwise default cases such as 500 will not be handled
+
+- The `attempt` property in the object callback refers to the current retriesAttempt number.
+You are given the freedom to use this in concert with the retryCount property set to handle retries
+
+- `getDefaultRetry` refers to the default retry handler. If this callback returns not null and you wish to handle those case of errors simply invoke it as done below.
+
+- `err`, `req`, `res` are self-explanatory error, request, and response properties respectively
```
airhorns, line 284:
```suggestion
- `err` is the error thrown by making a request using whichever agent is configured
- `req` is the raw request details sent to the underlying agent. __Note__: this object is not a Fastify request object, but instead the low-level request for the agent.
- `res` is the raw response returned by the underlying agent (if available) __Note__: this object is not a Fastify response, but instead the low-level response from the agent. This property may be null if no response was obtained at all, like from a connection reset or timeout.
```

index.js
```diff
@@ -142,36 +142,32 @@ const fastifyReplyFrom = fp(function from (fastify, opts, next) {
...
+      return null
+    }
 
-      requestImpl = createRequestRetry(request, this, retryHandler)
+    if (retryDelay && retryDelay.handler) {
+      requestImpl = createRequestRetry(request, this, (req, res, err, retries) => {
+        return retryDelay.handler({ err, req, res, attempt: retries, getDefaultDelay })
+      })
+    } else if (retryMethods.has(method) && !contentLength) {
+      requestImpl = createRequestRetry(request, this, getDefaultDelay)
```
airhorns, line 170: What about just always returning it the created retry-er, instead of duplicating the methods/content length check down here?

test/retry-with-a-custom-handler.test.js
```diff
@@ -59,79 +59,79 @@ test('a 500 status code with no custom handler should fail', async (t) => {
...
   }
 
-  const { instance } = await setupServer(t, { customRetry: { handler: customRetryLogic, retries: 10 } }, 500, 4, true)
+  const { instance } = await setupServer(t, { retryDelay: { handler: customRetryLogic } }, 500, 4, true)
 
-  const res = await got.get(`http://localhost:${instance.server.address().port}`, { retry: 0 })
+  const res = await got.get(`http://localhost:${instance.server.address().port}`, { retry: 5 })
 
   t.equal(res.headers['content-type'], 'text/plain')
   t.equal(res.statusCode, 205)
```
airhorns, line 137: I think a test that asserts on the `attempt` property and notes it goes up would be good, happy to pair on it if you like

index.js
airhorns: This has been changed to return an object that we can extend over time!

mcollina (approved):
lgtm

MikePresman:
Not sure why my above type definition fix would have failed the test. Could someone re run it? I'm unable to reproduce this on my macos machine

Uzlopak:
@MikePresman 

Did you address all remarks of @airhorns ? If so, could you comment on them accordingly and resolve them please?

Eomm (approved):

types/index.d.ts
```diff
@@ -39,12 +39,21 @@ declare module "fastify" {
 }
 
 type FastifyReplyFrom = FastifyPluginCallback<fastifyReplyFrom.FastifyReplyFromOptions>
-
 declare namespace fastifyReplyFrom {
   type QueryStringFunction = (search: string | undefined, reqUrl: string) => string;
+
+  export type customRetryHandler = {
```
airhorns, line 45: nit: This isn't really a handler so much as the details, no? 

```suggestion
  export type RetryDetails = {
```

Would need to update the readme too

types/index.d.ts
```diff
@@ -39,12 +39,21 @@ declare module "fastify" {
...
+    err: Error;
+    req: FastifyRequest<RequestGenericInterface, RawServerBase>;
+    res: FastifyReply<RawServerBase>;
+    attempt: number;
+    getDefaultDelay: () => number | null;
+  }
   export interface FastifyReplyFromHooks {
     queryString?: { [key: string]: unknown } | QueryStringFunction;
     contentType?: string;
+    retryDelay?: (retryObj: customRetryHandler) => {} | null;
```
airhorns, line 55:
```suggestion
    retryDelay?: (details: RetryDetails) => {} | null;
```

types/index.test-d.ts
```diff
@@ -4,7 +4,7 @@ import { IncomingHttpHeaders } from "http2";
 import * as https from 'https';
 import { AddressInfo } from "net";
 import { expectType } from 'tsd';
-import replyFrom, { FastifyReplyFromOptions } from "..";
+import replyFrom, { FastifyReplyFromOptions, customRetryHandler } from "..";
```
airhorns, line 7:
```suggestion
import replyFrom, { FastifyReplyFromOptions, RetryDetails } from "..";
```

types/index.test-d.ts
```diff
@@ -91,6 +91,15 @@ async function main() {
   instance.get("/http2", (request, reply) => {
       reply.from("/", {
           method: "POST",
+          retryDelay: ({err, req, res, attempt, getDefaultDelay}: customRetryHandler) => {
```
airhorns, line 94: Might be a better test to not explicitly type the object here and let the inference prove that its all working?

```suggestion
          retryDelay: ({err, req, res, attempt, getDefaultDelay}) => {
```

airhorns (approved):
LGTM after comments

Uzlopak (approved):
RSLGTM

types/index.d.ts
```diff
@@ -39,12 +39,21 @@ declare module "fastify" {
...
-
 declare namespace fastifyReplyFrom {
   type QueryStringFunction = (search: string | undefined, reqUrl: string) => string;
+
+  export type RetryDetails = {
+    err: Error;
+    req: FastifyRequest<RequestGenericInterface, RawServerBase>;
+    res: FastifyReply<RawServerBase>;
+    attempt: number;
+    getDefaultDelay: () => number | null;
```
Antiavanti, line 50: Is it valid signature? Shouldn't it be:
```
  export type DefaultRetryDetails = {
    err: Error;
    req: FastifyRequest<RequestGenericInterface, RawServerBase>;
    res: FastifyReply<RouteGenericInterface, RawServerBase>;
    retries: number;
  }
```
```
getDefaultDelay: (details: DefaultRetryDetails) => number | null;
```
Based on implementation:
https://github.com/fastify/fastify-reply-from/blob/cd2ba083e4a4e8085dc973edb8c5e622efaf830e/index.js#L151C3-L151C53
```
    const getDefaultDelay = (req, res, err, retries) => {
      if (retryMethods.has(method) && !contentLength) {
        // Magic number, so why not 42? We might want to make this configurable.
        let retryAfter = 42 * Math.random() * (retries + 1)

        if (res?.headers['retry-after']) {
          retryAfter = res.headers['retry-after']
        }
        if (res && res.statusCode === 503 && req.method === 'GET') {
          if (retriesCount === 0 && retries < maxRetriesOn503) {
            return retryAfter
          }
        } else if (retriesCount > retries && err && err.code === retryOnError) {
          return retryAfter
        }
      }
      return null
    }

    if (retryDelay) {
      requestImpl = createRequestRetry(request, this, (req, res, err, retries) => retryDelay({ err, req, res, attempt: retries, getDefaultDelay, retriesCount }))
    } else {
      requestImpl = createRequestRetry(request, this, getDefaultDelay)
    }
```
If yes, then also example from `README.md` is probably incorrect: https://github.com/fastify/fastify-reply-from/blob/cd2ba083e4a4e8085dc973edb8c5e622efaf830e/README.md?plain=1#L323

types/index.d.ts
```diff
@@ -39,12 +39,21 @@ declare module "fastify" {
...
+  }
   export interface FastifyReplyFromHooks {
     queryString?: { [key: string]: unknown } | QueryStringFunction;
     contentType?: string;
+    retryDelay?: (details: RetryDetails) => {} | null;
```
Antiavanti, line 55: Shouldn't  it be`retryDelay?: (details: RetryDetails) => number | null` (`number` instead of `{}`), the same as `getDefaultDelay` above (`number | null`), to be precise?

PR #359, opened 2024-04-03
Update for v5

gurgunday:


package.json
```diff
@@ -30,35 +30,35 @@
...
     "end-of-stream": "^1.4.4",
     "fast-content-type-parse": "^1.1.0",
-    "fast-querystring": "^1.0.0",
-    "fastify-plugin": "^4.0.0",
+    "fast-querystring": "^1.1.2",
+    "fastify-plugin": "^4.5.1",
     "pump": "^3.0.0",
     "toad-cache": "^3.7.0",
-    "undici": "^5.19.1"
+    "undici": "^5.28.3"
```
mcollina, line 61: Bump this to v6

mcollina (commented):
Can you also update tap?

I think I'll likely ship this sooner to expose the updated undici.

package.json
gurgunday: Will do both

gurgunday:
There are multiple errors when tap is updated to v18, and it doesn't seem trivial to fix them...

If people really need the release, we can follow-up later to upgrade `devDeps` separately

gurgunday:
Done

elmarcusec:
@gurgunday @mcollina I have a question. When and in which version do you plan to release changes to upgrade the undici library?

gurgunday:
This will be released as a major

gurgunday:
Ok I made a mistake, should've merged to next first 😄

Will fix in a sec

gurgunday:
This PR is ready, only note I'll leave is that for some reason, in certain cases `t.teardown` is not functioning and I have to resolve the promises myself, not sure if this is a tap issue or what

Fdawgs (approved):

PR #362, opened 2024-04-10
fix: DELETE requests should not call stream.end

karl-power:
It appears that `DELETE` streams behave the same as `GET` streams in that they are not writable (immediately ended?), so calling `.end` on the stream throws a `write after end` error ([related issue](https://github.com/fastify/fastify-http-proxy/issues/253)).

I cannot find this documented anywhere in HTTP2 spec or in NodeJS HTTP2 module (aside from the `endStream` header, and `http2stream.endAfterHeaders` which don't seem to affect anything here - can anyone share insight?).

This PR simply adds the `DELETE` method to a check on whether or not to end the stream. The change fixes https://github.com/fastify/fastify-http-proxy/issues/253.

#### Checklist

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -204,7 +204,8 @@ function buildRequest (opts) {
       ...stripHttp1ConnectionHeaders(opts.headers)
     }, http2Opts.requestOptions)
     const isGet = opts.method === 'GET' || opts.method === 'get'
-    if (!isGet) {
+    const isDelete = opts.method === 'DELETE' || opts.method === 'delete'
+    if (!isGet && !isDelete) {
       end(req, opts.body, done)
     }
     req.setTimeout(http2Opts.requestTimeout, () => {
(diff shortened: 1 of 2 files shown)


simoneb (approved):
Nice job. This change LGTM, but I'd like to wait for somebody else's review before merging.

gurgunday (approved):
Makes sense, lgtm

gurgunday:
> It appears that DELETE streams behave the same as GET streams in that they are not writable (immediately ended?), so calling .end on the stream throws a write after end error (https://github.com/fastify/fastify-http-proxy/issues/253).

Yeah I thought for the same reasons when approving, that DELETE requests are like GET requests, but I just learned that they actually [might](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE) have a body, so they can be writable (?) – maybe for all practical purposes it is interpreted the same way

Not 100% sure if this is the reason or not

karl-power:
Hi @mcollina & @climba03003
Any idea if this is the right approach here?

PR #363, opened 2024-04-15
feat: added option for undici proxy agent

### LINKED ISSUE #159: Support using undici behind a proxy server
## 🚀 Feature Proposal

Add a field in undici options that specifies the proxy you need to connect to before forwarding the request to the upstream.

## Motivation

Right now the only way to use the plugin behind a proxy server is to not use undici and instead use http while passing the agents option:

```ts
const ReplyFrom = require('fastify-reply-from');
const proxy = require('proxy-agent');
fastify.register(ReplyFrom, {
  http: {
     agents: {
       "http:": proxy('url'),
       "https:": proxy('url')
     }
  }
});
```

Recently undici has added support for proxies as mentioned in its [documentation](http://undici.nodejs.org/#/docs/best-practices/proxy), so it would be great if we can support passing the proxy field as part of undici's options and not have to switch to http.

## Example

```ts
const ReplyFrom = require('fastify-reply-from');
fastify.register(ReplyFrom, {
  unidici: {
    proxy: 'http://username:password@proxyIp:proxyPort'
  }
});
```

10xLaCroixDrinker:
Closes #159

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ proxy.register(require('@fastify/reply-from'), {
 
 #### `undici`
 
-By default, [undici](https://github.com/mcollina/undici) will be used to perform the HTTP/1.1
+By default, [undici](https://github.com/nodejs/undici) will be used to perform the HTTP/1.1
 requests. Enabling this flag should guarantee
 20-50% more throughput.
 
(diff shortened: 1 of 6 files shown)


mcollina (approved):
lgtm

Issue #366, opened 2024-04-26
Cannot completely disable requestTimeout and sessionTimeout


gunters63:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.26.2

### Plugin version

9.7.0

### Node.js version

v20.11.1

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

Ubuntu 20.04

### Description

I have an industrial application here which uses grpc-web in the browser to communicate with a .Net backend.
I use a fastify web server as a proxy to forward the HTTP/2 requests (content-type "application/grpc-web-text") to the .Net backend.

After some fiddling I got it to work with the help of @fastify/http-proxy (which uses @fastify/reply-from internally).

But I still have a somewhat exotic problem. The UI makes heavy use of Grpc subscriptions. A Grpc subscription keeps a HTTP/2 session open as long as the subscription is active.

The UI can sit idle for hours and days with the same active subscription(s), which means the default requestTimeout and sessionTimeout timeouts have to be disabled or increased.

Disabling is not possible because the code handles a timeout of 0 as undefined:

```
  if (!http2Opts.sessionTimeout) {
    http2Opts.sessionTimeout = opts.sessionTimeout || 60000
  }
  if (!http2Opts.requestTimeout) {
    http2Opts.requestTimeout = 10000
  }
```

I increased the timeouts to the maximal possible values instead (2147483647). Which means after 25 days the UI will fail if no one touches it. Which is unlikely but I don't want to rule it out.

In this use case it would best if you could completely disable the timeouts. So I propose to handle the 0 value explicitely in the code. I would also be happy to provide a merge request if this change is approved.

### Expected Behavior

The user should be able to completely disable requestTimeout and sessionTimeout by providing a value of 0.

gunters63:
Okidoki, I prepare one in the next days :)

gunters63:
@mcollina: I have prepared a draft pull request (https://github.com/fastify/fastify-reply-from/pull/368)

I wrote a test for the disabled request timeout and wanted it to fail with the original code and pass with my changes.
This troubled me a little, because the original code falls back to a default of 10000 ms if requestTimeout is falsy.
So currently the test has a long runtime of 11000ms. I am sure there is a better way, maybe you can give me hint.

[...]

PR #368, opened 2024-05-07
allow explicitly disabling request and session timeouts with 0

gunters63:
#### Checklist

### DIFF
--- a/test/http2-timeout-disabled.test.js
+++ b/test/http2-timeout-disabled.test.js
@@ -29,20 +29,14 @@ test('http2 request timeout disabled', async (t) => {
 
   await instance.listen({ port: 0 })
 
-  try {
-    await Promise.all([
-      got.get(`http://localhost:${instance.server.address().port}/`, {
-        retry: 0
-      }),
-      new Promise((resolve, reject) => setTimeout(reject, 11000, 'passed'))
-    ])
-  } catch (err) {
-    // if we wait 11000ms without a timeout error, we assume disabling the timeout worked
-    // 10000 ms is the default timeout
-    t.equal(err, 'passed')
+  const result = await Promise.race([
+    got.get(`http://localhost:${instance.server.address().port}/`, {
+      retry: 0
+    }),
+    new Promise(resolve => setTimeout(resolve, 11000, 'passed'))
+  ])
 
-    return
-  }
-
-  t.fail()
+  // if we wait 11000ms without a timeout error, we assume disabling the timeout worked
+  // 10000 ms is the default timeout
+  t.equal(result, 'passed')
 })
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

mcollina:
why is it still a draft?

gunters63:
I made a comment in the issue (https://github.com/fastify/fastify-reply-from/issues/366), I maybe should have better put it here.

The PR was still a draft because I was unsure about the missing test about disabling the sessionTimeout. I wrote a test, but it never completed and tap ran into a timeout. I don't really know what is the reason. So i left the test out.

Maybe this isn't a problem, so I'll undraft the PR now.

mcollina:
Ah I see it. So a test is missing for `requestTimeout`? It seems the one for `sessionTimeout` is there?

test/http2-timeout-disabled.test.js
```diff
@@ -0,0 +1,48 @@
...
+      }),
+      new Promise((resolve, reject) => setTimeout(reject, 11000, 'passed'))
+    ])
+  } catch (err) {
+    // if we wait 11000ms without a timeout error, we assume disabling the timeout worked
+    // 10000 ms is the default timeout
+    t.equal(err, 'passed')
+
+    return
+  }
```
Eomm, line 45:
```suggestion
    const result = await Promise.race([
      got.get(`http://localhost:${instance.server.address().port}/`, {
        retry: 0
      }),
      new Promise((resolve) => setTimeout(resolve, 11000, 'passed'))
    ])

    // if we wait 11000ms without a timeout error, we assume disabling the timeout worked
    // 10000 ms is the default timeout
    t.equal(result, 'passed')
```

gunters63:
> Ah I see it. So a test is missing for `requestTimeout`? It seems the one for `sessionTimeout` is there?

The other way around :)

test/http2-timeout-disabled.test.js
gunters63: Ah, yes, good point. race will work better here.

mcollina:
Can you add the test about sessionTimeout?

gunters63:
Ok, I think I solved the problem with the hanging test. I added a proper disabling session timeout test now.

I explicitly clean up the request and the two Fastify instances directly after the timeout. Then the test finishes cleanly. If I leave out any of this:

```js
  // clean up right after the timeout, otherwise test will hang
  request.cancel()
  target.close()
  instance.close()
```

the test hangs. Cleaning at teardown time (t.teardown) does not help here, maybe a chicken/egg problem.

mcollina (approved):
lgtm

PR #423, opened 2025-06-23
types: improve onResponse hook typing

### LINKED ISSUE #383: onResponse hook TypeScript definition issue with response.stream.
### Prerequisites

- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.0.0

### Plugin version

11.0.1

### Node.js version

20.x

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

14.5

### Description

Hello,

I am using the `onResponse` hook to properly implement some custom tracing with OpenTelemetry and need the hook to otherwise behave normally, but trigger some extra logic to clean up a OpenTelemetry Span stored in the request. Note that I am using this inside of `@fastify/http-proxy` but the underlying issue is coming from this package since that package simply wraps this one.

There seems to be a moderate issue when using TypeScript and overriding the `onResponse` hook. Specifically, there's seemingly no good way to access the `stream` on `RawReplyDefaultExpression<RawServerBase>` that's assigned to the `res` in that hook. I see that it's defined in Http2ServerResponse type in Fastify, and that `RawReplyDefaultExpression` is supposed to be a union of all the possible response types (HTTP and HTTP2). I have my proxy configured as HTTP/1.

Previously, I was able to get away with 

```javascript
onResponse: async (
        if (request.currentSubspan) {
          request.currentSubspan.setAttribute(
            "reply.elapsedTime",
            reply.elapsedTime
          );
          request.currentSubspan.end();
          request.currentSubspan = null;
        }
        reply.send(res);
),
http2: false,
```

But upon upgrading to Fastify 5.x and the latest version of this plugin, this no longer functions, despite `res` itself being defined as a stream. Using this approach currently results in a blank response where it previously.

If you try to use `res.stream`, you will get the following error in the TypeScript compiler.

```
Property 'stream' does not exist on type 'RawReplyDefaultExpression<RawServerBase>'.
  Property 'stream' does not exist on type 'ServerResponse<IncomingMessage>'.
```

Right now, my workaround is.

```javascript
      onResponse: async (
        request,
        reply,
        res: RawReplyDefaultExpression<RawServerBase>
      ) => {
        if (request.currentSubspan) {
          request.currentSubspan.setAttribute(
            "reply.elapsedTime",
            reply.elapsedTime
          );
          request.currentSubspan.end();
          request.currentSubspan = null;
        }
        // eslint-disable-next-line @typescript-eslint/no-explicit-any
        reply.send((res as any)?.stream);
      },
      http2: false,
```

This is really suboptimal for pretty obvious reasons. The actual structure of `response` is not reflected in the TypeScript definition, so as a result, it's difficult to use properly. I suppose the "fix" here would be to ensure there's a definition of `stream` in the type definition.

### Link to code that reproduces the bug

_No response_

### Expected Behavior

I would expect the `response.stream` property to be defined in some fashion in the exported TypeScript definitions for the `onResponse` hook.

t-tajiri:
#### Summary
This Pull Request is fix for https://github.com/fastify/fastify-reply-from/issues/383.

#### Checklist

- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -15,6 +15,7 @@ import {
   Agent,
   AgentOptions,
   IncomingHttpHeaders,
+  IncomingMessage,
   RequestOptions,
 } from 'node:http'
 import {
@@ -55,6 +56,11 @@ declare namespace fastifyReplyFrom {
     retriesCount: number;
     getDefaultDelay: () => number | null;
   }
+
+  export type RawServerResponse<T extends RawServerBase> = RawReplyDefaultExpression<T> & {
+    stream: IncomingMessage
+  }
+
   export interface FastifyReplyFromHooks {
     queryString?: { [key: string]: unknown } | QueryStringFunction;
     contentType?: string;
@@ -63,7 +69,7 @@ declare namespace fastifyReplyFrom {
     onResponse?: (
       request: FastifyRequest<RequestGenericInterface, RawServerBase>,
       reply: FastifyReply<RouteGenericInterface, RawServerBase>,
-      res: RawReplyDefaultExpression<RawServerBase>
+      res: RawServerResponse<RawServerBase>
     ) => void;
     onError?: (
       reply: FastifyReply<RouteGenericInterface, RawServerBase>,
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #438, opened 2025-09-18
Add test for text/event-stream proxying with custom parser

mcollina:
Adds test and documentation for proxying text/event-stream using custom content type parser.


gurgunday (approved):
lgtm

